PR #68, opened 2025-07-07
Add NEL-related attributes

armenzg:
The changes are reflecting the discussions from [this document](https://www.notion.so/sentry/Browser-Reports-Attribute-21e8b10e4b5d8045a7fdc936c814ed70) and follow-up to [this Relay PR](https://github.com/getsentry/relay/pull/4874).

### DIFF
--- a/model/attributes/browser/browser__report__type.json
+++ b/model/attributes/browser/browser__report__type.json
@@ -1,10 +1,10 @@
 {
-    "key": "browser.report.type",
-    "brief": "A browser report sent via reporting API..",
-    "type": "string",
-    "pii": {
-        "key": "false"
-    },
-    "is_in_otel": false,
-    "example": "network-error"
-}  
+  "key": "browser.report.type",
+  "brief": "A browser report sent via reporting API..",
+  "type": "string",
+  "pii": {
+    "key": "false"
+  },
+  "is_in_otel": false,
+  "example": "network-error"
+}
--- a/model/attributes/browser/browser__report__type.json
+++ b/model/attributes/browser/browser__report__type.json
@@ -1,0 +1,10 @@
+{
+    "key": "browser.report.type",
+    "brief": "A browser report sent via reporting API..",
+    "type": "string",
+    "pii": {
+        "key": "false"
+    },
+    "is_in_otel": false,
+    "example": "network-error"
+}  
--- a/model/attributes/nel/nel__elapsed_time.json
+++ b/model/attributes/nel/nel__elapsed_time.json
@@ -1,7 +1,7 @@
 {
   "key": "nel.elapsed_time",
   "brief": "The elapsed number of milliseconds between the start of the resource fetch and when it was completed or aborted by the user agent.",
-  "type": "number",
+  "type": "integer",
   "pii": { "key": "false" },
   "is_in_otel": false,
   "example": 100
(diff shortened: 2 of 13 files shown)


model/attributes/browser/browser__report__type.json
```diff
@@ -0,0 +1,10 @@
+{
+    "key": "browser.report.type",
+    "brief": "A browser report sent via reporting API..",
```
armenzg, line 3:
```suggestion
    "brief": "A browser report sent via reporting API.",
```

model/attributes/nel/nel__elapsed_time.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "nel.elapsed_time",
+    "brief": "The elapsed number of milliseconds between the start of the resource fetch and when it was completed or aborted by the user agent.",
+    "type": "number",
```
armenzg, line 4: Is this a valid type? Should it be `int`?

model/attributes/nel/nel__phase.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "nel.phase",
+    "brief": "If request failed, the phase of its network error. If request succeeded, \"application\".",
+    "type": "string",
+    "pii": { "key": "false" },
+    "is_in_otel": false,
+    "example": "application"
```
armenzg, line 7: e.g. dns, application , connection

model/attributes/nel/nel__referrer.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "nel.referrer",
```
armenzg, line 2: Intentional double `rr` 😄

model/attributes/nel/nel__sampling_function.json
```diff
@@ -0,0 +1,8 @@
+{
+    "key": "nel.sampling_function",
+    "brief": "The sampling function used to determine if the request should be sampled.",
+    "type": "float",
```
armenzg, line 4: Is `float` a valid type?

model/attributes/nel/nel__elapsed_time.json
```diff
@@ -0,0 +1,8 @@
...
+    "pii": { "key": "false" },
+    "is_in_otel": false,
+    "example": 100
```
armenzg, line 7: For numeric examples, I assume I can put it without quotes.

model/attributes/nel/nel__elapsed_time.json
AbhiPrasad: right now we only have `number`: https://github.com/getsentry/sentry-conventions/blob/253b600b1eddbfc98dd02684440738ee877c4f6f/schemas/attribute.schema.json#L51

but we should probably expand it to be `int` vs. `float`. I'll make the change.

model/attributes/nel/nel__sampling_function.json
AbhiPrasad: Only `number` for now.

model/attributes/nel/nel__elapsed_time.json
AbhiPrasad: yup, this is valid. The tests should catch this.

PR #71, opened 2025-07-08
CI: Fail if generate produces files

armenzg:
For now, this will not commit any files to reduce the amount of work here.


armenzg:
bugbot run

cleptric:
We will likely only want to do this on a new release.

AbhiPrasad (approved):

PR #78, opened 2025-07-28
feat(otlp): Add first set of convention documentation for span names

gggritso:
This is an experimental PR! I'm soliciting suggestions on how to tackle this, and this PR is meant as a point of discussion to figure out how to move forward.

---

👋🏻 I would like to start documenting semantic conventions for the `name` field of a span. To me, semantic conventions for `name` are useful in at least two places:

1. SDKs. As we create more and more V2 SDKs (or continue to use POTel) we will need to set a `name` on every span. Since Sentry makes spans that OTel does not, we need to document conventions for how to make span names, so the SDKs create the span names consistently. I've already noticed inconsistencies in casing of some `name` attributes in the AI namespace
2. In the product. There's at least one use case in Relay where we need to generate a `name` attribute from a span's other attributes and fields. We need a way to refer to what span names should look like, rather than hunting through every SDK

I added a schema for a span name definition, added some examples, and added a simple documentation page generator. Here are the main approaches of my first take:

- The files have the same directory structures as `attributes`, roughly. There is one file for every directory in `attributes`
- Each name definition has a few fields:
  - A brief description (not that useful IMO, but nice to look at)
  - A list of span `op`s that are relevant. The idea is that this documentation is relevant for only a few known `op`s. I assume this is contentious since we're moving away from `op` fields, but this is critical to know in the code, since we need to know which span to apply the conventions to. Perhaps `category` would be more palatable?
  - A list of templates on how to create a span name from attributes. This would be great to codegen, but it also works well as documentation, IMO, combined with the examples. I'm a little worried that not every case can be done via this kind of template approach, but I haven't kicked the tires on it yet
- There is a little script to create a single documentation page for span name generation. Maybe separate pages are better, but I wanted to spike something quick for us to talk about

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,21 @@ After you've created an attribute, the script will ask if you'd like to generate
 
 If you need help, run `yarn run create:attribute --help` to see the available options.
 
+## Adding a new convention for span names
+
+Span name conventions are organized loosely by type of span operation. To create a convention for a new type of span operation:
+
+1. Create a new file in the `models/name` directory. Ideally the file name should match a folder in the `models/attributes` directory (e.g., `ui.json`).
+2. Fill in the contents of the file with the necessary information. You can find the schema for the document in `schemas/name.schema.json`.
+
+- The `"brief"` field should be a short description of what kind of information the `name` field contains for this kind of span operation.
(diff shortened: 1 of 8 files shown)


AbhiPrasad (approved):
Thank you!

PR #98, opened 2025-08-07
feat: Add conventions for internal span buffer attributes

gggritso:
Right now, we have a loose convention to prefix attributes with `__sentry_internal`to hide those attributes from the UI. We want to improve this a bit with a friendlier convention of using the prefix `sentry.internal`. This PR documents two of the known existing ones just to get the process started.


AbhiPrasad (approved):
It's a little annoying this doesn't follow the conventions to use a period, but otherwise lgtm!

gggritso:
@AbhiPrasad @cleptric I ended up namespacing them under `sentry.internal.` rather than `__sentry_internal` or even `sentry._internal`. The repo has a _lot_ of feelings about where underscores should and should not go, and I didn't want to futz with that too much. What do you think?

cleptric:
This is because we haven't added `sentry._meta` yet 😬

cleptric (approved):

PR #105, opened 2025-08-27
ref: Make `replacement` non-required on `attribute.deprecation`

lcian:
### Commit 1
Move duplicated `AttributeJson` type to a `common.ts` file

### Commit 2
Certain attributes are completely deprecated in OTEL, see https://opentelemetry.io/blog/2023/http-conventions-declared-stable/#:~:text=as%20server.port-,net.sock.peer.name,-Removed
We were representing this as `"replacement": ""`, which is incorrect.
This change makes the `replacement` field on attributes optional to be able to handle cases like this, adapts the generation logic, and updates the relevant generated files by rerunning the `generate` script.

Other examples of completely removed OTEL attributes https://opentelemetry.io/docs/specs/semconv/non-normative/db-migration/

### Alternative
As an alternative, we could instead remove the problematic attribute from the conventions when we consider something to be completely removed, but I think it's better to represent it as deprecated without any replacement.

### DIFF
--- a/generated/attributes/net.md
+++ b/generated/attributes/net.md
@@ -189,8 +189,8 @@ Peer address of the network connection - Unix domain socket name
 | Has PII | false |
 | Exists in OpenTelemetry | Yes |
 | Example | `/var/my.sock` |
-| Deprecated | Yes, use `` instead |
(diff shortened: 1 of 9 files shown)


lcian:
We can address this in a different way.

PR #121, opened 2025-09-04
feat: Generate Python code for attributes

lcian:
Exposes 2 big constants:
- `ATTRIBUTE_NAMES`: class containing a key for each attribute in the registry, mapping the constant to the attribute name. 
Raises a warning when accessing deprecated names
Meant to be used in the SDK/product code.
- `_ATTRIBUTE_METADATA`: dict mapping each attribute name to its metadata. 
Typed as a `dict[str, AttributeMetadata]` so it's possible to try to access it with arbitrary keys and it will return `None` if that attribute is not in the conventions.
Meant to be used in the buffer.

Usage example:
```python
from sentry_conventions import attributes

attr_str = "ai.responses"

metadata = attributes._ATTRIBUTE_METADATA.get(attr_str)
ty = metadata.type if metadata else None
print(ty)  # attributes.AttributeType.STRING_ARRAY

deprecation = metadata.deprecation if metadata else None
replacement = deprecation.replacement if deprecation else None
print(replacement)  # "gen_ai.response.text"


attr_str = attributes.ATTRIBUTE_NAMES.HTTP_RESPONSE_CONTENT_LENGTH  # emits a DeprecationWarning

metadata = attributes._ATTRIBUTE_METADATA.get(attr_str)
deprecation = metadata.deprecation if metadata else None
status = deprecation.status if deprecation else None
print(status)  # attributes.DeprecationStatus.BACKFILL


attr_str = "not.in.conventions"

metadata = attributes._ATTRIBUTE_METADATA.get(attr_str)
print(metadata)  # None
```

### DIFF
--- a/python/src/sentry_conventions/attributes.py
+++ b/python/src/sentry_conventions/attributes.py
@@ -153,6 +153,8 @@ class _AttributeNamesMeta(type):
         "RELEASE",
         "REPLAY_ID",
         "ROUTE",
+        "SENTRY_BROWSER_NAME",
+        "SENTRY_BROWSER_VERSION",
         "_SENTRY_SEGMENT_ID",
         "TRANSACTION",
         "URL",
@@ -569,6 +571,7 @@ class ATTRIBUTE_NAMES(metaclass=_AttributeNamesMeta):
     Type: str
     Contains PII: maybe
     Defined in OTEL: No
+    Aliases: sentry.browser.name
     Example: "Chrome"
     """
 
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

untitaker (commented):
how is this code going to be used? from what i can tell mostly in queries? you might want to generate a typeddict with these keys, as well as mapping AttributeType to actual python types that you'd expect back from the query interface:

```
Attributes = TypedDict({"ai.citations": list[str]})
```

i'd suggest to try to introduce this code pre-release into sentry itself to validate that it is useful as-is + adding CI in this repo to validate that the generated file doesn't crash on import.

otherwise lgtm

lcian:
Added two TypedDicts: `Attributes` and `FullAttributes`, including the deprecated ones.
We might want to just use `FullAttributes` in place of `Attributes` and remove the latter, because with the current setup every time we deprecate something we'd remove it from the `TypedDict` which would be a breaking change.

PR #123, opened 2025-09-05
chore(ai): Remove unused gen_ai cost attributes

vgrozdanic:
We don't want to report reasoning or cached tokens as a special cost parameters for now, so this will be removed from the conventions.

Part of [TET-1100: Update relay and conventions to reduce work scattering](https://linear.app/getsentry/issue/TET-1100/update-relay-and-conventions-to-reduce-work-scattering)


obostjancic (approved):

lcian (approved):
By the way now there's also the option to completely deprecate something (by not specifying a replacement).
That way it would remain in the repo/generated files for posterity.
However if you're sure this is not used anywhere at all right now, I guess removing it completely to reduce confusion is better.

PR #128, opened 2025-09-09
dynamic-sampling: add field conventions for dynamic sampling context

shellmayr:
- Add dynamic sampling context to sentry-conventions for storage when JSON store is removed

Contributes to TET-887


cleptric:
We likely want to filter those out in the UI, so maybe we can prefix them with `sentry._internal`?

cleptric (approved):

PR #137, opened 2025-09-11
feat(sentry): Add sentry.observed_timestamp_nanos

loewenheim:
This attribute is [added by Relay](https://github.com/getsentry/relay/blob/add4bfc8a16bbac41d32cb0d34cd4d844ac3bf1c/relay-event-normalization/src/eap/mod.rs#L61-L71) based on the time an envelope is received.

Closes INGEST-573.


loewenheim:
I've added the attribute both with and without the `_internal` segment now, the latter is deprecated.

lcian (approved):

PR #155, opened 2025-09-24
chore(ai): Clean up of `sentry._internal.segment.contains_gen_ai_spans`

vgrozdanic:
Code was cleaned up in https://github.com/getsentry/sentry/pull/99993, we decided not to use it as we found other workarounds to make it simpler.


cleptric:
As long as no SDK has ever emitted it, we don't break self-hosted (Relay Pii scrubbing) and no more usage in the product we can.

shellmayr (approved):
LGTM from the product point view 👍

cleptric (approved):

PR #163, opened 2025-10-03
feat: Add vercel log drain attributes

AbhiPrasad:
resolves https://linear.app/getsentry/issue/LOGS-388/document-vercel-log-drain-attributes-in-conventions

This PR adds a set of attributes to the conventions for attributes that we'll generate in the upcoming vercel log drain endpoint.

These attributes are created when we map the vercel log schema to the sentry schema and attributes as relevant. See this notion doc for details about the schema and the transform: https://www.notion.so/sentry/Vercel-Log-Drain-2808b10e4b5d808c95f3ebcfe2b8828a.

Most attribute keys here align in naming as per https://vercel.com/docs/drains/reference/logs. There are some fields in the vercel log drain schema that don't show up in this PR. These are:

1. `timestamp`, which is mapped directly to the log timestamp
2. `level` which is mapped directly to the log level
3. `message` which is mapped directly to the log body

### DIFF
--- a/generated/attributes/index.md
+++ b/generated/attributes/index.md
@@ -48,3 +48,4 @@ This directory contains documentation for all available attributes.
 - [`url` Attributes](./url.md)
 - [`user` Attributes](./user.md)
 - [`user_agent` Attributes](./user_agent.md)
+- [`vercel` Attributes](./vercel.md)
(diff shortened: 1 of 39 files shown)


lcian (approved):
LGTM

PR #171, opened 2025-10-16
feat(mcp): MCP attributes

betegon:
Adding MCP attributes.


python/src/sentry_conventions/attributes.py
```diff
@@ -5,12 +5,9 @@
 import warnings
```
betegon: ok not really sure about the Python attributes. @constantinius is working on them, so they are not really added yet. This was generated automatically, Should I leave it as it is?

betegon:
Fixed `is_in_otel` and PII `false` attributes, @lcian could you please have a quick look at it again?

lcian (approved):
Thanks @betegon, looks good.

PR #174, opened 2025-10-20
ref: Undeprecate `sentry.observed_timestamp_nanos`

loewenheim:
## Description
https://github.com/getsentry/sentry-conventions/pull/137 added `sentry.observed_timestamp_nanos` as a deprecated attribute and `sentry._internal.observed_timestamp_nanos` as its non-deprecated replacement. However, adding the attribute as deprecated was probably a mistake because it is in active use and no concrete plan was made for a migration to the `_internal` version. It's also questionable whether this attribute should be hidden from users in the first place.

Since the `_internal` version is not used anywhere and not written by Relay, I propose that we delete it and mark the non-`_internal` version as not deprecated.

## PR Checklist


Dav1dde:
>  That we display sentry.observed_timestamp_nanos in the product is more a bug than a feature, how should any user know what this even means?

I think it's just on a hover and renamed to something reasonable.

This just highlights a fundamental issue, maybe we shouldn't tie to the visibility (frontend state) to the attribute name, but instead track that separately as metadata for an attribute. If for changing the visibility state we have to change ingestion and backfilll for 90 days, that seems pretty bad.

This also came up recently, where we may want to track frontend aliasing also in the conventions.

PR #176, opened 2025-10-21
fix: Rename paths that aren't valid on Windows

loewenheim:
## Description
In addition to renaming files containing `<,>`, this also runs tests on Windows to make sure such invalid names are caught.

## PR Checklist

#skip-changelog

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -2,7 +2,7 @@ name: Danger
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


lcian (approved):
Good catch, thanks!
Hmm I see the build is failing, let me look into that.
Ah alright, it seems that https://github.com/getsentry/sentry-conventions/pull/177 is intended to fix this.

PR #177, opened 2025-10-21
fix(generation): Make order deterministic

loewenheim:
## Description
JS code generation relies on non-deprecated attributes being visited before deprecated ones for constant name generation, but this was never enforced. This causes tests to fail under Windows, where files are apparently listed in a different order.

Therefore, this PR makes the iteration order in JS code generation deterministic by sorting attributes by:
1. What name the constant should have;
2. Deprecation (non-deprecated first).

Python generation is unchanged; the problem isn't so critical there because constant names have already been generated and memoized at that point.

## PR Checklist

### DIFF
--- a/scripts/generate_attributes.ts
+++ b/scripts/generate_attributes.ts
@@ -46,7 +46,7 @@ function writeToJs(attributesDir: string, attributeFiles: string[]) {
   usedConstantNames.clear();
 
   // First pass: collect all attributes
-  var allAttributesPartial: Array<{
+  const allAttributesPartial: Array<{
     file: string;
     key: string;
     attributeJson: AttributeJson;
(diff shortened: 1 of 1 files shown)


scripts/generate_attributes.ts
```diff
@@ -57,11 +57,49 @@ function writeToJs(attributesDir: string, attributeFiles: string[]) {
...
+      return 1;
+    } else {
+      if (a.isDeprecated === b.isDeprecated) {
+        return 0;
+      } else if (a.isDeprecated) {
+        return 1;
+      } else {
+        return -1;
+      }
+    }
```
loewenheim: Someone who is good at JavaScript please help me write this sorting function

lcian:
Thanks for taking care of this @loewenheim, I thought about doing this previously but at least on macos the files were always sorted in the same way, so I didn't think this would be an issue.
I've added some changes on top of this in https://github.com/getsentry/sentry-conventions/pull/178 to satisfy the linter, feel free to incorporate them (I think you should be able to merge that branch into this one).

lcian (approved):
LGTM, I guess we can skip changelog for this one as it's an internal change in the codegen.

PR #183, opened 2025-11-03
feat: Add timber logging attributes

markushi:
Add support for Timber logging framework attributes, specifically the timber.tag attribute for capturing log tags from the Timber logging framework in Sentry Android SDK.


markushi:
> yarn generate && yarn format

@lcian  Yeah I did that. Does it look better now? Which files should be there and which shouldn't?

PR #187, opened 2025-11-04
chore: Add .envrc and pre-commit hook

lcian:
## Description

Adds a pre-commit hook to run the formatting and linting steps.
In addition, adds a `.envrc` to run `yarn install` and install the pre-commit hook.
I believe a global uv should be available on all machines thanks to devenv, but please @cleptric or @AbhiPrasad give this a try locally and let me know if that's really the case. Otherwise we will need a devenv config too.


cleptric:
I don’t have uv installed fwiw 😅

PR #194, opened 2025-11-19
feat(attributes): Add sentry.normalized_db_query

Ahmed-Labs:
## Description

Introduces a new sentry attribute `sentry.normalized_db_query`. In the V1 span Relay pipeline, we read the raw db query from the span description field, scrub it, and write it back to the description field (which becomes `sentry.normalized_description`). 

Now that we're migrating to V2 spans, we plan on reading the query from the conventional field `db.query.text`, normalizing it, and writing it to another field (instead of overwriting the raw original field). However, there is currently no appropriate field to use for the normalized query.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## Unreleased
+
+### Features
+
+#### Changes to attributes
+
+- feat(attributes): Add sentry.normalized_db_query ([#194](https://github.com/getsentry/sentry-conventions/pull/194))
+
 ## 0.3.0
 
 ### Features


mjq:
@cleptric 

> We might want to mark this as deprecated then and add a back-fill property to let Relay write it into the correct field.

`sentry.normalized_description` currently holds several different types of values, depending on the type of span (`op`). So, this isn't a 1:1 replacement. We'll similarly move the other types of values into their own semantic attributes.

cleptric:
Ok, I think I understand. So `db.query.text`, while it should be already sanitized (https://opentelemetry.io/docs/specs/semconv/database/database-spans/#sanitization-of-dbquerytext), might not be, so we sanitize it again and use `sentry.normalized_db_query` in the product. And our SDKs will send `db.query.text`.

PR #196, opened 2025-11-21
feat(otel): write OTel environment into `sentry.environment`

mjq:
## Description
By convention, OTel SDKs sends environment name in the `deployment.environment.name` resource attribute ([spec](https://opentelemetry.io/docs/specs/semconv/resource/deployment-environment/)).

We would like Sentry's environment filters to work with OTel data. This PR attempts to use sentry-conventions' backfilling to accomplish this, writing the incoming data to our own conventional `sentry.environment` attribute.

During ingestion, resource attributes are flattened into our single attributes hash with the keys prefixed with `resource.`, so during Relay's processing pipeline, the full attribute name of the source attribute `resource.deployment.environment.name`. That is the attribute name defined in this PR.

Mechanically this is exactly what I want, but semantically it feels a little weird. Comments below.

Related to DAIN-1099.

## PR Checklist

If an attribute was added:

- [ ] I've followed the policies described in [CONTRIBUTING.md](https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Sentry's semantic conventions align with [OpenTelemetry's semantic conventions](
 
 These will be a standardized naming scheme for operations and data that will be shared across the SDKs, ingest, and the product. This allows us to have a single source of truth for the semantic conventions, and also allows us to generate code for all parts of the stack (ingest, frontend, backend, data pipelines) that need to be aware of this.
 
-Conventions may also map attributes from OpenTelemetry resource attributes into Sentry's semantic attributes. Such OpenTelemetry resousrce attributes are ingested into Sentry as regular attributes prefixed with `resource.`. For example, see definitions in [the `model/attributes/resource` directory](model/attributes/resource).
+Conventions may also map attributes from OpenTelemetry resource attributes into Sentry's semantic attributes. Such OpenTelemetry resource attributes are ingested into Sentry as regular attributes prefixed with `resource.`. For example, see definitions in [the `model/attributes/resource` directory](model/attributes/resource).
 
 To see the current list of attributes, see [the generated documentation for attributes](generated/attributes/index.md).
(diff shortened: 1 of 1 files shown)


model/attributes/resource/resource__deployment__environment__name.json
```diff
@@ -0,0 +1,14 @@
+{
+  "key": "resource.deployment.environment.name",
+  "brief": "The software deployment environment name.",
+  "type": "string",
+  "pii": {
+    "key": "false"
+  },
+  "is_in_otel": true,
```
mjq: `resource.deployment.environment.name` isn't in OTel per se, but `deployment.environment.name` does exist as a resource attribute. 😕

AbhiPrasad: We can probably treat `resource.X` attributes as a special case. Maybe worth leaving a note about this in the README.

model/attributes/resource/resource__deployment__environment__name.json
```diff
@@ -0,0 +1,14 @@
...
+  "example": "production",
+  "deprecation": {
+    "_status": "backfill",
+    "replacement": "sentry.environment"
+  }
```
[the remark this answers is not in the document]
AbhiPrasad: Maybe instead of deprecation we introduce a new field for this. I don't want to block this PR though, so we can maybe add a note to the brief and create a follow up issue.

model/attributes/resource/resource__deployment__environment__name.json
mjq: Annoyingly, we already have a `resource.X` attribute that has nothing to do with OTel resources. Still, I tried adding a note to the readme.

PR #199, opened 2025-12-02
chore(attributes): Normalize deprecated db attribute names

Ahmed-Labs:
Normalizes deprecated db attributes. It's important that these attributes get their names normalized to the latest convention because we normalize their values afterwards (and we expect the latest attribute names during normalization).

### DIFF
--- a/model/attributes/db/db__operation.json
+++ b/model/attributes/db/db__operation.json
@@ -8,7 +8,7 @@
   "is_in_otel": true,
   "example": "SELECT",
   "deprecation": {
-    "_status": "backfill",
+    "_status": "normalize",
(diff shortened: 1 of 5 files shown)


lcian:
> Backfills deprecated db attributes. It's important that these attributes get their names normalized to the latest convention because we normalize their values afterwards (and we expect the latest attribute names during normalization).

So you always want to read the value from the non-deprecated name, right?
Then what you want is `normalize`, not `backfill`.

lcian (approved):
Before merging, please ensure that nothing relies on the old attribute names, as with this change we won't write to those anymore. There might be code paths that still read from them.

PR #206, opened 2025-12-10
chore(ai): Remove gen_ai_usage_total_cost attribute

vgrozdanic:
Part of [TET-1099: Stop double writing cost and update the product to query the new attribute](https://linear.app/getsentry/issue/TET-1099/stop-double-writing-cost-and-update-the-product-to-query-the-new)


shellmayr (approved):
Nice 🚀

PR #208, opened 2025-12-15
fix(db): Clarify `db.query.text` and `db.query.summary` attributes

Lms24:
* `db.query.text`: We previously specified this attribute to contain a full/raw query without parameterization. OTel requires the [paramterized query](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-query-text).
* `db.query.summary` We specified this attribute to contain the paramterized query. [OTel specifies](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-query-summary) a shortened operation/table "grouping".

This came up while reviewing https://github.com/getsentry/sentry-javascript/pull/17961

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+## Unreleased
+
+### Features
+
+- Add page listing all attributes ([#207](https://github.com/getsentry/sentry-conventions/pull/207))
+
+### Fixes
+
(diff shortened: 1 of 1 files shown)


mjq:
Yeah this is fine. We're already set up to use `.text` for insights. The product has no plans for using `.summary` (relying on our own parameterization and normalization instead), but if we want to document it because POtel SDKs send it, that's cool.

PR #237, opened 2026-01-23
feat(attributes): Add `browser.connection.rtt` attribute

Lms24:
Our JS SDK records a `connection.rtt` measurement, measuring the estimated request RTT based on the current connection quality. This is not really a performance indicator of made requests, but rather just an estimate how fast a request/response cycle will likely be handled. Anyway, we record this on the pageload and navigation spans (not sure about the value tbh but we do). 

For converting this measurement to an attribute, I for now decided to use the `browser` namespace due to this value being a coarse estimation.  We could use the `network` namespace and add something like `network.connection.rtt`. Just not sure if we want these estimations connected to the more general attribute. But tbh, LOGAF-L, so whatever reviewers prefer.

### DIFF
--- a/generated/attributes/all.md
+++ b/generated/attributes/all.md
@@ -4,7 +4,7 @@
 
 This page lists all available attributes across all categories.
 
-Total attributes: 423
+Total attributes: 425
 
 ## Stable Attributes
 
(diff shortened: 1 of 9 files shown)


mjq (approved):
`network.connection` [already exists as a namespace in OTel's semconv](https://opentelemetry.io/docs/specs/semconv/general/attributes/#network-connection-and-carrier-attributes) and seems to fit for this use too so I would vote for `network.connection.rtt`, but I also don't feel too strongly about it.

PR #261, opened 2026-02-20
fix(gen_ai): input and output token description

obostjancic:
## Description
- closes [TET-1986: Fix input and output token descriptions in sentry conventions](https://linear.app/getsentry/issue/TET-1986/fix-input-and-output-token-descriptions-in-sentry-conventions)

## PR Checklist

If an attribute was added:

- [x] I've followed the policies described in [CONTRIBUTING.md](https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md)


shellmayr (approved):
LGTM from my side 👍

PR #266, opened 2026-02-24
feat(resource): add resource.deployment.environment

mjq:
## Description
We currently have a definition for `resource.deployment.environment.name`, deprecated for `sentry.environment` with a `backfill` policy. That means that incoming spans using the OpenTelemetry conventional attribute `deployment.environment.name` in their span's `resource` will get the value copied to `sentry.environment` in Relay, which makes the environment filters work across the product. (see https://github.com/getsentry/sentry-conventions/pull/196)

`deployment.environment.name` was [introduced in OTel's semantic conventions v1.27.0 (Aug 2024)](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.27.0) as a replacement for `deployment.environment`, which was deprecated at that time. One of our customers is still using this deprecated attribute, and I'm sure they're not alone in that.  Define `resource.deployment.environment` as a deprecated, backfilled attribute so that clients using it, too will get working environment filters in the product.

## PR Checklist

If an attribute was added:
    - `false` to match `sentry.environment` (presumably because the product depends on the exact given value)

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -6273,6 +6273,27 @@ export const REPLAY_ID = 'replay_id';
  */
 export type REPLAY_ID_TYPE = string;
 
+// Path: model/attributes/resource/resource__deployment__environment.json
+
+/**
+ * The software deployment environment name. `resource.deployment.environment`
+ *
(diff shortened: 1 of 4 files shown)


jflayhart:
thanks y'all!

PR #127, opened 2025-09-08
ref: Set PII to maybe for most non `sentry.*` attributes

lcian:
### Summary of changes

Set to `pii = "true"`:
- `http.client_ip`
- and its replacement `client.address`

Left to `pii: "false"` or things could break:
- id: "A unique identifier for the span."
- profile_id
- replay_id
- type: "More granular type of the operation happening." e.g. "fetch"

Notable attributes changed to `pii: "maybe"`:
- transaction
- release
- environment: "The Sentry environment."

All of the remaining attributes have also been changed from PII `false` to `maybe`.
Some of these are not typed at all in Relay (which means `pii: true`), others are typed and have `pii: false`/`maybe`. I document the current value in Relay for some of them below.
Note that all of the `pii: "true"` cases are due to the fact that the attribute is not typed in Relay, explicit `pii: "true"` values are present only on SentryTags.

- ai.ai__finish_reason - Relay: false
- ai.ai__generation_id - Relay: false
- ai.ai__metadata - Relay: false
- ai.ai__model__provider - Relay: false
- ai.ai__model_id - Relay: true
- ai.ai__pipeline__name - Relay: false
- ai.ai__response_format - Relay: true
- ai.ai__responses - Relay: maybe
- ai.ai__seed - Relay: false
- ai.ai__tags - Relay: true
- ai.ai__tools - Relay: maybe
- app_start_type - Relay: maybe (in SentryTags)
- browser.browser__name - Relay: maybe (typed in SentryTags)
- browser.browser__report__type - Relay: maybe
- browser.browser__script__invoker - Relay: maybe
- browser.browser__script__invoker_type - Relay: maybe
- browser.browser__version - Relay: false
- cache.cache__operation - Relay: true
- channel - Relay: true
- client.client__address - Relay: false
- code.code__function__name - Relay: maybe
- db.db__collection__name - Relay: false
- db.db__name - Relay: true
- db.db__namespace - Relay: true
- db.db__operation - Relay: false
- db.db__operation__name - Relay: false
- db.db__query__parameter__[key] - Relay: true
- db.db__query__summary - Relay: true
- db.db__query__text - Relay: true
- db.db__redis__connection - Relay: true
- db.db__statement - Relay: true
- db.db__system - Relay: false
- db.db__system__name - Relay: true
- device.device__brand - Relay: maybe (typed in SentryTags)
- device.device__family - Relay: maybe (SentryTags)
- device.device__model - Relay: maybe (SentryTags)
- error.error__type - Relay: true
[...]

PR #270, opened 2026-02-27
feat: Add changelog tracking to attribute definitions

### LINKED ISSUE #268: Display a changelog/history of attributes on the Sentry Conventions docs website
What I'm envisioning is something along the lines of:

```
Some attribute
---
Version 1.0: Introduced
Version 1.1: Calculation of property now includes cached tokens where previously it didn't. Conforms to changes in OTEL spec as of v3.x.x
```

For example, the `gen_ai.cost.input_tokens` [attribute](https://getsentry.github.io/sentry-conventions/attributes/#gen_ai-cost-input_tokens)  previously excluded cached input tokens and now includes them.

When the OTel spec [was updated last month](https://github.com/open-telemetry/semantic-conventions/commit/9d7d97ac117127662dafe3fd3094d5d660d6f3b7) to reflect this, [our conventions were changed](https://github.com/getsentry/sentry-conventions/commit/6fe20412d93bdc5f856656469726cee5be8c7008) to follow, but the Sentry SDK versions that are older would still be calculating this value with the previous definition, which can lead to confusion for users who visit the docs to better understand the metrics generated in Sentry.

ericapisani:
Fixes #268 and CON-82

Add a `changelog` field to attribute definitions that tracks how each attribute has changed across sentry-conventions release versions.

This makes it possible to show users when an attribute was introduced, deprecated, or changed — which is useful context when browsing the attribute docs or consuming the metadata in SDKs.

## What's included

- **Schema** (`schemas/attribute.schema.json`): New optional `changelog` array with `version`, `prs`, and `description` fields
- **Generation script** (`scripts/generate_attribute_changelog.ts`): Derives changelog entries from git tags and commit history, extracting PR numbers from commit messages
- **Code generators** (`scripts/generate_attributes.ts`): Propagates `changelog` into the generated TypeScript (`ChangelogEntry` interface on `AttributeMetadata`) and Python (`ChangelogEntry` dataclass) outputs
- **Docs UI** (`docs/src/components/AttributeCard.astro`): Renders a collapsible "Changelog" section on each attribute card, linking to PRs on GitHub
- **Tests** (`test/attributes.test.ts`): Validates changelog entry structure and ordering (newest first)
- **All attribute JSON files**: Populated with generated changelog entries based on git history

## Screenshots

Screenshot 2026-02-27 at 2 04 10 PM

Screenshot 2026-02-27 at 2 04 25 PM

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -9846,7 +9846,10 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     isInOtel: false,
     example: ['Citation 1', 'Citation 2'],
     deprecation: {},
-    changelog: [{ version: '0.1.0', prs: [55] }],
+    changelog: [
+      { version: 'next', prs: [264] },
+      { version: '0.1.0', prs: [55] },
+    ],
   },
   [AI_COMPLETION_TOKENS_USED]: {
     brief: 'The number of tokens used to respond to the message.',
@@ -9872,7 +9875,10 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     isInOtel: false,
     example: ['document1.txt', 'document2.pdf'],
     deprecation: {},
-    changelog: [{ version: '0.1.0', prs: [55] }],
+    changelog: [
+      { version: 'next', prs: [264] },
+      { version: '0.1.0', prs: [55] },
+    ],
   },
   [AI_FINISH_REASON]: {
     brief: 'The reason why the model stopped generating.',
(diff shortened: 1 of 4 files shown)


scripts/generate_attribute_changelog.ts
```diff
@@ -0,0 +1,147 @@
...
+    const filePath = path.join(attributesDir, relativeFile);
+    const gitPath = path.join('model', 'attributes', relativeFile);
+    const changelog = buildChangelog(gitPath, tagRanges);
+
+    if (changelog.length === 0) {
+      continue;
+    }
+
+    const attributeJson = JSON.parse(fs.readFileSync(filePath, 'utf-8')) as AttributeJson;
+    attributeJson.changelog = changelog;
```
Lms24: m: does this overwrite pre-existing changelog entries? I'm not sure if I'm missing something but it seems like in the future, we might want to run this script again for PRs that touch a lot of files (like #127) but there's also value in manually writing a changelog entry (for example, if we change semantics of an attribute like in #208 and add a description to the changelog). Would we overwrite then manually added entry?

Lms24:
> Yes, but I can look into modifying this so that it preserves pre-existing entries (including manually written ones that you mentioned in another comment), while still handling the edge case of a transition where what was listed under a next version is correctly updated to the latest version number.

sounds great, thanks!

> When you say "producing changes", are you referring to something like a diff existing between what's in the PR vs. what the script generated?

Yes basically. I checked our Build CI job and it seems like [we already fail](https://github.com/getsentry/sentry-conventions/blob/5e5df001c3a2b282b52fce3dad9e75e74f459ce8/.github/workflows/build.yml#L44-L57) if `yarn generate` causes any changes on CI. So if we run the changelog script on `yarn generate` we should be good!

PR #288, opened 2026-03-16
docs(sentry): Add deprecated `sentry.source`

s1gr1d:
Adds an already existing (and deprecated) attribute `sentry.source` and also adds additional information about the distinction between `route` and `url`.

### DIFF
--- a/model/attributes/sentry/sentry__source.json
+++ b/model/attributes/sentry/sentry__source.json
@@ -14,7 +14,7 @@
   },
   "changelog": [
     {
-      "version": "0.0.0"
+      "version": "next"
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


model/attributes/sentry/sentry__source.json
```diff
@@ -0,0 +1,20 @@
...
+  "is_in_otel": false,
+  "example": "route",
+  "deprecation": {
+    "_status": "backfill",
+    "replacement": "sentry.span.source",
+    "reason": "This attribute is being deprecated in favor of sentry.span.source"
+  },
+  "changelog": [
+    {
+      "version": "0.0.0"
```
Lms24: l: this correlates to the version of the conventions package. So `next` for now as it's unreleased.
```suggestion
      "version": "next"
```

PR #295, opened 2026-03-19
feat(attributes): Add `db.driver.name` attribute

### LINKED ISSUE #294: Add `db.driver.name` attribute
Document attribute proposed in https://github.com/getsentry/sentry-python/issues/5463

alexander-alderman-webb:
## Description

An idea proposed by the community in https://github.com/getsentry/sentry-python/issues/5463

Closes https://github.com/getsentry/sentry-conventions/issues/294

## PR Checklist

If an attribute was added:

- [ ] I've followed the policies described in [CONTRIBUTING.md](https://github.com/getsentry/sentry-python/blob/master/CONTRIBUTING.md)


Lms24 (approved):
Sounds good to me! Looks like the only remotely similar OTel attribute is [`db.jdbc.driver_classname`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-jdbc-driver-classname) but it's deprecated and way too specific. I think it makes sense to have a more general attribute to record the driver.

PR #308, opened 2026-04-02
feat(attributes): add `gen_ai.function_id` attribute

constantinius:
## Description

Add the `gen_ai.function_id` attribute. Currently only emitted by Vercel AI in the JS SDK.

Closes https://linear.app/getsentry/issue/TET-2181/add-gen-aifunction-id-to-sentry-conventions

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -3029,7 +3029,7 @@ export type GEN_AI_EMBEDDINGS_INPUT_TYPE = string;
 // Path: model/attributes/gen_ai/gen_ai__function_id.json
 
 /**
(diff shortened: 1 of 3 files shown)


alexander-alderman-webb (changes requested):

alexander-alderman-webb:
There are other occurrences of the description that need to be updated.
I think that's why CI is failing @constantinius

constantinius:
> There are other occurrences of the description that need to be updated. I think that's why CI is failing @constantinius

@alexander-alderman-webb fixed

alexander-alderman-webb (approved):

Lms24 (approved):

PR #313, opened 2026-04-08
feat(attributes): Add app.vitals.* attributes for mobile performance

buenaflor:
Add new `app.vitals.*` namespaced attributes for mobile app performance monitoring and deprecate their old counterparts.

**New attributes:**
- `app.vitals.start.cold.value` — cold app start duration in milliseconds
- `app.vitals.start.warm.value` — warm app start duration in milliseconds
- `app.vitals.start.type` — app start type (cold/warm)
- `app.vitals.ttid.value` — time to initial display in milliseconds
- `app.vitals.ttfd.value` — time to full display in milliseconds
- `app.vitals.frames.total.count` — total frames rendered during span lifetime
- `app.vitals.frames.slow.count` — slow frames rendered during span lifetime
- `app.vitals.frames.frozen.count` — frozen frames rendered during span lifetime
- `app.vitals.frames.delay.value` — sum of delayed frame durations in seconds

**Deprecated (backfill status, with aliases on both sides):**
- `app_start_type` → `app.vitals.start.type`
- `time_to_initial_display` → `app.vitals.ttid.value`
- `time_to_full_display` → `app.vitals.ttfd.value`
- `frames.total` → `app.vitals.frames.total.count`
- `frames.slow` → `app.vitals.frames.slow.count`
- `frames.frozen` → `app.vitals.frames.frozen.count`
- `frames.delay` → `app.vitals.frames.delay.value`


model/attributes/app/app__vitals__frames__delay__value.json
```diff
@@ -0,0 +1,19 @@
+{
+  "key": "app.frames.delay.value",
+  "brief": "The sum of all delayed frame durations in seconds during the lifetime of the span. For more information see [frames delay](https://develop.sentry.dev/sdk/performance/frames-delay/).",
+  "type": "integer",
```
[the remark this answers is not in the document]
buenaflor: wrong, sdks send integers specifically

model/attributes/app/app__ttfd__value.json
```diff
@@ -0,0 +1,19 @@
+{
+  "key": "app.ttfd.value",
```
Lms24: l/Q: For browser, I decided to name-space former web vital measurements to `browser.web_vital.<shortVitalName>.value`. I'm not fully sure around the semantics of TTFD and TTID but do we show them as "[Mobile Vitals](https://docs.sentry.io/product/insights/mobile/mobile-vitals/)" in the product?

Just in case we do: Have you thought about naming them `app.vitals.ttfd.value` (and ttid respectively)? I wouldn't put a hard requirement on this though and will happily leave it up to you. So feel free to disregard! I'm probably missing a lot of semantics here. 

(also, I don't know which of the attributes in this PR we consider "vitals" vs. other, so please feel free to extend or reduce this feedback to whatever makes sense)

model/attributes/app/app__frames__delay__value.json
```diff
@@ -0,0 +1,19 @@
+{
+  "key": "app.frames.delay.value",
```
Lms24: l/q: Why do we need the `.value` here? 

(feel free to disregard. Might also be related to my other point about vitals)

Lms24 (approved):
Had two optional suggestions/questions but otherwise LGTM!

model/attributes/app/app__frames__delay__value.json
buenaflor: mainly future proofing here in case we want to add more data in the future (no real reason other than that)

model/attributes/app/app__ttfd__value.json
buenaflor: hm yeah I think it makes sense to put them under vitals as well<br><br>all of the added attributes here are considered to be mobile vitals so I think we should

Lms24 (approved):
thanks!

PR #317, opened 2026-04-13
feat(attributes): Add `sentry.mobile` and `sentry.main_thread` attributes

buenaflor:
## Description

Add two Sentry attributes (note these are also pre-computed by relay in v1 transactions and used today in transactions):

- **`sentry.mobile`** — Whether a mobile SDK is being used.
- **`sentry.main_thread`** — Whether the event occurred on the main thread.

Both attributes are computed by Relay and should not be set by SDKs.

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -7983,7 +7983,7 @@ export type SENTRY_KIND_TYPE = string;
  *
  * Attribute Value Type: `boolean` {@link SENTRY_MAIN_THREAD_TYPE}
  *
- * Contains PII: maybe
+ * Contains PII: false
  *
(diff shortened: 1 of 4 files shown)


Lms24 (approved):
LGTM, wondering though if we should add a flag to the attribute schema that we can use to determine which component in the stack (SDK, Relay, Product) is expected to set the attribute. Probably warrants greater discussion though so feel free to just merge this as-is.

buenaflor:
> wondering though if we should add a flag to the attribute schema that we can use to determine which component in the stack (SDK, Relay, Product) is expected to set the attribute

agreed! filed an issue: https://github.com/getsentry/sentry-conventions/issues/320

PR #319, opened 2026-04-14
feat(attributes): Add LCP and CLS web vital report event attribute

Lms24:
Two web vital attributes were missing which we should still include for web vital meta information:

- `browser.web_vital.lcp.report_event`
- `browser.web_vital.cls.report_event`

These attributes carry information on what event caused the SDK to report a web vital span for CLS and LCP: Either a `pagehide` or `navigation` event. More events could follow in the future.

For both of these web vitals, we previously sent `sentry.report_event` but in retrospect I believe this was a bad naming choice (my fault lol). Given we have a common pattern now for web vitals, let's align these. No product logic depends on these attributes. They're merely meta information.

ref: 
- https://github.com/getsentry/sentry-javascript/pull/19827
- https://github.com/getsentry/sentry-javascript/issues/17931

### DIFF
--- a/model/attributes/sentry/sentry__report_event.json
+++ b/model/attributes/sentry/sentry__report_event.json
@@ -1,0 +1,24 @@
+{
+  "key": "sentry.report_event",
+  "brief": "(Deprecated) The event that caused the SDK to report CLS or LCP (pagehide or navigation)",
+  "type": "string",
+  "pii": {
+    "key": "maybe"
+  },
+  "is_in_otel": false,
+  "example": "pagehide",
+  "alias": ["browser.web_vital.lcp.report_event", "browser.web_vital.cls.report_event"],
+  "sdks": ["javascript-browser"],
+  "deprecation": {
+    "replacement": null,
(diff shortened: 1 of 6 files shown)


model/attributes/sentry/sentry__report_event.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "is_in_otel": false,
+  "example": "pagehide",
+  "alias": ["browser.web_vital.lcp.report_event", "browser.web_vital.cls.report_event"],
+  "sdks": ["javascript-browser"],
+  "deprecation": {
+    "replacement": null,
+    "reason": "The report event is now recorded as a browser.web_vital.lcp.report_event or browser.web_vital.cls.report_event attribute. No backfill required.",
+    "_status": null
+  },
```
Lms24, line 14: I documented the old attribute (currently still sent with v1 standalone web vital spans) and marked it as deprecated. Since there's no 1:1 replacement but SDKs send one of two different attributes depending on the span, I left `repalcement` and `_status` intentionally on `null` since I don't think we can backfill this at all. Given there's no product logic around these attributes I think that's fine but happy to do something else if reviewers have concerns.

cleptric (approved):

PR #336, opened 2026-04-23
feat(attributes): Add http.request.body.data attribute

ericapisani:
Add a new HTTP attribute `http.request.body.data` capturing the request body payload — JSON body or form data, with any attached files redacted.

Defined as a `string` with `pii: maybe` since request bodies can carry sensitive user data, and `is_in_otel: false` because the OpenTelemetry semantic conventions registry does not define an equivalent. Placing it under the existing `http.request.*` namespace keeps it discoverable alongside `http.request.header.<key>` and `http.response.body.size`.

Model JSON is the source of truth; the TypeScript and Python attribute files were regenerated via `yarn generate` and should not be hand-edited.

Refs PY-2362

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -5322,7 +5322,7 @@ export type HTTP_QUERY_TYPE = string;
 // Path: model/attributes/http/http__request__body__data.json
 
 /**
- * HTTP request body data. Could be the JSON body or the form data. If form data, and a file is attached, the file is redacted. `http.request.body.data`
+ * HTTP request body data. Can be given as string or structural data of any format. `http.request.body.data`
  *
  * Attribute Value Type: `string` {@link HTTP_REQUEST_BODY_DATA_TYPE}
  *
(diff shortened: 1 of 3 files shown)


javascript/sentry-conventions/src/attributes.ts
```diff
@@ -15965,6 +15987,17 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     example: '?foo=bar&bar=baz',
     changelog: [{ version: '0.0.0' }],
   },
+  [HTTP_REQUEST_BODY_DATA]: {
+    brief:
+      'HTTP request body data. Could be the JSON body or the form data. If form data, and a file is attached, the file is redacted.',
```
cleptric: Today we document it as https://develop.sentry.dev/sdk/foundations/transport/event-payloads/request/#attributes

ericapisani: Addressed in https://github.com/getsentry/sentry-conventions/pull/336/commits/e247c74e6a1b957ed996685adc8849a0f6df0e28

cleptric (approved):

sentrivana (approved):

PR #337, opened 2026-04-23
feat(attributes): Add middleware.name attribute

ericapisani:
Add a new span attribute `middleware.name` for capturing the name of middlewares, client or server side.

Refs PY-2362

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -7308,6 +7308,26 @@ export const METHOD = 'method';
  */
 export type METHOD_TYPE = string;
 
+// Path: model/attributes/middleware/middleware__name.json
+
+/**
+ * The name of the middleware. `middleware.name`
+ *
+ * Attribute Value Type: `string` {@link MIDDLEWARE_NAME_TYPE}
+ *
+ * Contains PII: false
+ *
+ * Attribute defined in OTEL: No
+ *
+ * @example "AuthenticationMiddleware"
+ */
+export const MIDDLEWARE_NAME = 'middleware.name';
+
+/**
+ * Type for {@link MIDDLEWARE_NAME} middleware.name
+ */
+export type MIDDLEWARE_NAME_TYPE = string;
+
 // Path: model/attributes/navigation/navigation__type.json
 
 /**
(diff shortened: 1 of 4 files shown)


cleptric (approved):

javascript/sentry-conventions/src/attributes.ts
```diff
@@ -9938,6 +9938,26 @@ export const SERVICE_VERSION = 'service.version';
...
+ *
+ * Attribute Value Type: `string` {@link STARLETTE_MIDDLEWARE_NAME_TYPE}
+ *
+ * Contains PII: false
+ *
+ * Attribute defined in OTEL: No
+ *
+ * @example "AuthenticationMiddleware"
+ */
+export const STARLETTE_MIDDLEWARE_NAME = 'starlette.middleware.name';
```
alexander-alderman-webb: There are other similar data/tags in the SDK, e.g., `litestar.middelware_name`:

https://github.com/getsentry/sentry-python/blob/1faec55e6edb8bdafa46b3be4f7a20e0e055d000/sentry_sdk/integrations/litestar.py#L165

Have you considered making a more generic attribute? Something like `http.middleware.name`?
Having a more generic attribute could allow us to avoid defining many similar attributes.

ericapisani: Yeah, we can definitely explore making it more generic. I was aiming to maintain existing naming conventions with this change, but I guess given this is going to be a part of a major release (span-first), we can break with that.

Given middleware is not part of the HTTP protocol I'm hesitant to place it under `http` within the conventions, but the two approaches that I think would work better are:
- `middleware.name`
  - top-level, can be used by both server- and client-side middleware
- `server.middleware.name`/`client.middleware.name`
  - nested under the type of service in which they'd run, but results in duplication of attributes that we'd have to maintain.

I'm leaning more towards `middleware.name` but keen to know what others think.

cleptric: `middleware.name` is fine.

ericapisani: Great! Addressed in [53934b4](https://github.com/getsentry/sentry-conventions/pull/337/commits/53934b479906260349a2e7a3423e75a454e76241), marking this thread as resolved.

sentrivana (approved):
🙏🏻

PR #345, opened 2026-04-29
feat(attributes): Deprecate sentry.transaction

buenaflor:
Deprecate `sentry.transaction` in favor of `sentry.segment.name`.

The segment name is the preferred attribute going forward, so this marks `sentry.transaction` for backfill-based deprecation and updates the existing `transaction` alias to point at the same replacement.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -9822,6 +9822,8 @@ export type _SENTRY_SEGMENT_ID_TYPE = string;
  *
  * Attribute defined in OTEL: No
  *
+ * Aliases: {@link SENTRY_TRANSACTION} `sentry.transaction`, {@link TRANSACTION} `transaction`
+ *
  * @example "GET /user"
  */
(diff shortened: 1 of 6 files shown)


mjq:
I agree in principle but this one needs _extremely careful checking_ across the board before merge please - IIRC currently both this and `sentry.segment.name` are in use but in different places, and there's a pretty high possibility something breaks.

buenaflor:
> I agree in principle but this one needs extremely careful checking across the board before merge please - IIRC currently both this and sentry.segment.name are in use but in different places, and there's a pretty high possibility something breaks.

yeah definitely, I'll let this one sit for a while and investigate a bit, it's not critical

I noticed some dashboards are currently not working with span v2 since they require the "transaction" attribute (which is not backfilled or coalesced without the deprecation)

mjq:
@buenaflor Gotcha, that makes sense. I'm going on vacation for a bit but I bet @jameskeane can help sort out dashboard team help for the investigation if that's useful.

buenaflor:
I double checked and found no potential issues in deprecating it in favour of sentry.segment.name

cleptric (approved):

PR #347, opened 2026-04-30
feat(attributes): Add `process.runtime.engine.*` attributes

chargome:
## Description
Adds two new attributes for defining the js engine name + version.

ref https://github.com/getsentry/sentry-javascript/issues/20381

## PR Checklist

If an attribute was added:


cleptric (commented):
I think we should make this a bit more generic, like browser engine version. Nothing in Otel for that?

chargome:
nothing in otel, but we could do `js.engine.name` and `js.engine.version`?

Or `process.runtime.engine.name`.

nicohrubec (approved):

cleptric (approved):

PR #348, opened 2026-04-30
feat(attributes): Add `typescript.version` attribute

chargome:
## Description
Adds a new attr for ts versioning. Though about moving this into some kind of `build.*` namespace but not sure if we need that.

## PR Checklist

If an attribute was added:


cleptric:
hmm, might worth dropping. Like does this add any valuable info to your telemetry?

chargome:
TBH not much. Would just give us more insights on which ts versions users are on.

chargome:
not gonna happen, we'll drop it

PR #353, opened 2026-05-05
feat(attributes): Add app vitals start attributes

buenaflor:
Add `app.vitals.start.screen` and `app.vitals.start.reason` for app start spans.

The screen attribute records the first interactive screen after launch, while the reason attribute records the normalized launch trigger such as `push`. Omitting the screen attribute indicates a background app start where no UI was rendered.

Generated the TypeScript and Python convention outputs from the new attribute models.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -1160,7 +1160,7 @@ export type APP_VITALS_START_COLD_VALUE_TYPE = number;
 // Path: model/attributes/app/app__vitals__start__reason.json
 
 /**
- * The reason that triggered the app start. Values may differ between iOS and Android. `app.vitals.start.reason`
+ * The reason that triggered the app start. `app.vitals.start.reason`
  *
  * Attribute Value Type: `string` {@link APP_VITALS_START_REASON_TYPE}
  *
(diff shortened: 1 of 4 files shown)


cleptric (approved):

javascript/sentry-conventions/src/attributes.ts
```diff
@@ -13668,6 +13712,29 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
...
   },
+  [APP_VITALS_START_REASON]: {
+    brief: 'The reason that triggered the app start.',
+    type: 'string',
+    pii: {
+      isPii: 'maybe',
+    },
+    isInOtel: false,
+    example: 'push',
+    sdks: ['sentry.cocoa', 'sentry.java.android', 'sentry.javascript.react-native', 'sentry.dart.flutter'],
```
romtsn: dunno if we need to mention maui here?

romtsn (approved):
LGTM

itaybre (approved):
LGTM

javascript/sentry-conventions/src/attributes.ts
buenaflor: this sdks field is not really used anywhere, only informational

added it here as well though to be correct

PR #358, opened 2026-05-11
feat(attributes): add sentry.dsc.project_id as an attribute

constantinius:
## Description
Required for Dynamic Sampling

Contributes to https://linear.app/getsentry/issue/TET-2341/write-root-transaction-and-project-onto-every-span-in-relay

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -9266,7 +9266,7 @@ export type SENTRY_DSC_RELEASE_TYPE = string;
 // Path: model/attributes/sentry/sentry__dsc__root_project.json
 
 /**
- * The root project ID from the dynamic sampling context. `sentry.dsc.root_project`
+ * The ID of the project where the trace originated (i.e. the project of the SDK that started the trace). Propagated through the dynamic sampling context and set by Relay during ingestion. `sentry.dsc.root_project`
  *
  * Attribute Value Type: `string` {@link SENTRY_DSC_ROOT_PROJECT_TYPE}
  *
@@ -18661,7 +18661,8 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     changelog: [{ version: '0.3.0', prs: [185] }],
   },
   [SENTRY_DSC_ROOT_PROJECT]: {
-    brief: 'The root project ID from the dynamic sampling context.',
+    brief:
+      'The ID of the project where the trace originated (i.e. the project of the SDK that started the trace). Propagated through the dynamic sampling context and set by Relay during ingestion.',
     type: 'string',
     pii: {
       isPii: 'false',
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -9263,6 +9263,26 @@ export const SENTRY_DSC_RELEASE = 'sentry.dsc.release';
  */
 export type SENTRY_DSC_RELEASE_TYPE = string;
 
+// Path: model/attributes/sentry/sentry__dsc__root_project.json
+
+/**
+ * The root project ID from the dynamic sampling context. `sentry.dsc.root_project`
+ *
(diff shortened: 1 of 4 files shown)


javascript/sentry-conventions/src/attributes.ts
```diff
@@ -18638,6 +18660,16 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     example: 'frontend@e8211be71b214afab5b85de4b4c54be3714952bb',
     changelog: [{ version: '0.3.0', prs: [185] }],
   },
+  [SENTRY_DSC_ROOT_PROJECT]: {
+    brief: 'The root project ID from the dynamic sampling context.',
```
cleptric: Is this something relay/ingestion puts into the DSC? If so, worth documenting.

constantinius: For now this will be filled by Relay, but in the future this could also be set in the SDKs themselves (when we create a new DSC we can also set the root project, I believe, and then it is just passed on)

cleptric: How would we know the project in the SDK? The DSN contains it, but there are likely edge-cases.

constantinius: Isn't this part of the DSN? (maybe that is just my misunderstanding)

cleptric: Yes, but not always.

javascript/sentry-conventions/src/attributes.ts
```diff
@@ -9263,6 +9263,26 @@ export const SENTRY_DSC_RELEASE = 'sentry.dsc.release';
...
+ *
+ * Attribute Value Type: `string` {@link SENTRY_DSC_ROOT_PROJECT_TYPE}
+ *
+ * Contains PII: false
+ *
+ * Attribute defined in OTEL: No
+ *
+ * @example "12345"
+ */
+export const SENTRY_DSC_ROOT_PROJECT = 'sentry.dsc.root_project';
```
shellmayr: Is this the _best_ name for this attribute? @Dav1dde you might have some ideas about naming here, I think maybe we have a convention in relay already. What I know is that we use this as `count_per_root_project` in generic metrics, but maybe we can come up with a clearer name.

python/src/sentry_conventions/attributes.py
```diff
@@ -12443,6 +12455,20 @@ class ATTRIBUTE_NAMES(metaclass=_AttributeNamesMeta):
             ChangelogEntry(version="0.3.0", prs=[185]),
         ],
     ),
+    "sentry.dsc.root_project": AttributeMetadata(
+        brief="The root project ID from the dynamic sampling context.",
```
shellmayr: We should probably explain here what this attribute contains, i.e. that this is the project that a trace originates from.

javascript/sentry-conventions/src/attributes.ts
constantinius: Fixed in the latest version

python/src/sentry_conventions/attributes.py
constantinius: Added more details in the brief

javascript/sentry-conventions/src/attributes.ts
Dav1dde: I think `project_id` would fit, since we already have `public_key` which is also scoped to the DSC and therefor always the "root" -> the root is implied through the DSC.

But this is pretty confusing if one isn't familiar with DS(C)

constantinius: If the namespace `dsc` is enough to distinguish it then I'm fine with that. `sentry.dsc.transaction` also denotes the trace root transaction, so I guess this is fine here. Will change it to that.

constantinius: @Dav1dde done and also changed project ID type to `integer`, hope thats fine

Dav1dde (approved):

PR #363, opened 2026-05-12
feat(attributes): Add `profile.start_timestamp` attribute

chargome:
## Summary
- Add `profile.start_timestamp` (double) attribute for recording when a profile started
- Needed by the JS SDK (JS-2221) for span streaming, where scope contexts are converted to segment span attributes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,4 @@ dist
 .playwright-mcp/
 
 .DS_Store
+.claude/projects/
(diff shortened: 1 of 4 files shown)


cleptric:
Is this needed? Not mentioned on https://develop.sentry.dev/sdk/telemetry/profiles/.

chargome:
It is JS-specific, I'll drop it for now

PR #365, opened 2026-05-12
feat(attributes): Add `state.type` attribute

chargome:
## Summary
- Add `state.type` (string) attribute for identifying the state management library (e.g. "redux", "pinia")
- Needed by the JS SDK (JS-2221) for span streaming


loewenheim (commented):
Is this attribute's name already being used somewhere? It seems a bit generic.

chargome:
The JS SDK has been sending this for a while already yes

loewenheim (approved):

loewenheim:
In that case, approved :)

PR #366, opened 2026-05-12
feat(attributes): Add `missing_instrumentation.*` attributes

chargome:
## Summary
- Add `missing_instrumentation.package` (string) — the name of the uninstrumented package
- Add `missing_instrumentation.javascript.is_cjs` (boolean) — whether the package was loaded as CommonJS
- Used by the JS SDK to detect missing OTel instrumentation packages, needed for span streaming (JS-2221)


chargome:
I think this is not really useful on spans tbh, let's skip it for now.

PR #377, opened 2026-05-14
feat: Add span name conventions for MCP server spans

### LINKED ISSUE #376: Add conventions for names of MCP server spans
Right now there are no documented conventions for MCP spans. This means that span name inference wouldn't work correctly, and we have no reference for what the correct name would be. This already caused us to make incorrect decisions when making pre-built MCP dashboards which rely on the incorrect name `"mcp.server"`.

gggritso:
We've been sorely missing a convention for how to make MCP server span names! This documents them, since I just ran into an issue. OTel has some newer guidelines here (e.g., they use `gen_ai.tool.name`) but I'm sticking to our current conventions, with the template lifted out of Python SDK.

Downstream, this _would_ have affected how `name` is inferred for MCP server spans, but at this point we just use the transaction name, so it doesn't matter.

Closes #376.


betegon (approved):
thanks @gggritso! we'll catch up with OTEL and update things, but for now, let's stick to our convention as it's how the mcp sdk from anthropic works right now.

Lms24 (approved):
LGTM from conventions PoV. I'll defer to MCP integration authors for semantic correctness (@betegon already approved anyway, so I assume this is fine) and implementation (and the decision when to follow this).  From my PoV we should document the ideal naming in conventions anyway. 

Can we make sure this also correlates with what we document in the [develop specification](https://develop.sentry.dev/sdk/foundations/client/integrations/mcp/tracing/)? Or rather, I'd link to conventions for the naming instead of having it in both places.

cleptric:
Shall we move forward with this?

gggritso:
@mjq any objections to this? As far as I can tell, this doesn't have any downstream effect because MCP transaction spans use the transaction name as the span name anyway, and span name inference doesn't run. This PR mostly just documents what I saw already exists in the Python SDK

mjq (approved):

PR #385, opened 2026-05-18
feat(attributes): Add several deprecated measurements attributes

loewenheim:
## Description
This adds some attributes with the same names as existing measurements. The attributes are added as deprecated because they all already have updated names. The reason for adding them is because Relay converts measurements to attributes verbatim and it needs to access these attributes by name for calculations.

ref: INGEST-909

## PR Checklist

If an attribute was added:


loewenheim:
Closing this in favor of #388.

PR #388, opened 2026-05-19
feat(measurements): Add attribute field

### LINKED ISSUE #386: Allow declaring replacement attribute names in measurements

loewenheim:
## Description
This adds an optional field `"attribute"` of type `string` to the measurements schema. This field is meant to contain the name of the attribute corresponding to this measurement, if any. All existing measurements with corresponding attributes have also had this field filled in.

It also tests that if an attribute is set for a measurement, that attribute exists and does not have a replacement. The intention is that if a measurement's current attribute gets deprecated in favor of a new one, the measurement should be updated to point to the replacement attribute.

The purpose of this change is primarily to allow Relay to be smarter when converting measurements to attributes.

Closes #386. Closes CON-94.

## PR Checklist

### DIFF
--- a/model/measurements/app_start_cold.json
+++ b/model/measurements/app_start_cold.json
@@ -3,5 +3,6 @@
   "full_name": "App Start Cold",
   "brief": "The time it takes for the app to start from a cold state",
   "unit": "millisecond",
-  "platform": "mobile"
+  "platform": "mobile",
+  "attribute": "app.vitals.start.cold.value"
 }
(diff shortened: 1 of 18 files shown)


Lms24 (approved):
Nice, thanks! FYI, I'm currently working on adding a docs page for measurements so that we can surface them better. For Relay, do you need anything in terms of code generation for measurements?

Dav1dde (approved):

loewenheim:
> For Relay, do you need anything in terms of code generation for measurements?

We'll make the remaining changes in Relay itself. Currently it doesn't use measurements at all.

PR #390, opened 2026-05-19
feat(attributes): Add graphql.document attribute

ericapisani:
Add the `graphql.document` span attribute to capture the GraphQL document being executed.

This new attribute is part of the GraphQL semantic conventions and allows instrumentation to capture the full GraphQL operation string. Since documents may contain sensitive information in arguments or variables, instrumentation implementations should redact sensitive values when possible.

Includes attribute definitions across JavaScript, Python, and the model specification.

Refs #2366

### DIFF
--- a/model/attributes/graphql/graphql__document.json
+++ b/model/attributes/graphql/graphql__document.json
@@ -10,7 +10,8 @@
   "example": "query findBookById { bookById(id: ?) { name } }",
   "changelog": [
     {
-      "version": "0.0.0"
+      "version": "next",
+      "description": "Adds the `graphql.document` attribute to track the GraphQL document being executed."
     }
   ]
 }
--- a/model/attributes/graphql/graphql__document.json
+++ b/model/attributes/graphql/graphql__document.json
@@ -1,0 +1,16 @@
+{
+  "key": "graphql.document",
+  "brief": "The GraphQL document being executed.",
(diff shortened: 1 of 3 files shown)


model/attributes/graphql/graphql__document.json
```diff
@@ -0,0 +1,16 @@
...
+  "type": "string",
+  "pii": {
+    "key": "maybe",
+    "reason": "The document may contain sensitive information in arguments or variables. Instrumentation should redact sensitive information when possible."
+  },
+  "is_in_otel": true,
+  "example": "query findBookById { bookById(id: ?) { name } }",
+  "changelog": [
+    {
+      "version": "0.0.0"
```
mjq: This should be `"version": "next"`, ideally with `"prs": [390]` and a quick `"description"` too 🙏

ericapisani: Good catch, have fixed it in [0f8110a](https://github.com/getsentry/sentry-conventions/pull/390/commits/0f8110ac1be1f71c7cd25bcdc0bf7a0ff552e4a9)!

mjq (approved):
@ericapisani LGTM, but you'll have to run `yarn generate` to update all the client libraries before it'll let you merge.

sentrivana (approved):

PR #391, opened 2026-05-19
feat(attributes): Ensure replacements aren't deprecated

mjq:
When an attribute is deprecated, it should point to a non-deprecated replacement.

This should obviously be true for newly introduced deprecations, but is a tradeoff for the case where:
- attribute A is deprecated and replaced by B
- later, attribute B is deprecated and replaced by C

This would now cause a test failure, and A would need to also be updated to point to C instead of B.

However, on the plus side:
- much easier to programatically translate or replace attribute with their latest replacement, since you don't have to follow a chain of attribute definitions
- avoids the trap where a human sees a deprecated attribute and replaces it with the advertised replacement, not realizing that it's _also_ deprecated

I think on the whole this is worth it.

---

🤖: Claude (Opus 4.6) used to write the test; validated by hand. All words my own.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -170,7 +170,7 @@ export type AI_GENERATION_ID_TYPE = string;
  *
  * Aliases: {@link GEN_AI_REQUEST_MESSAGES} `gen_ai.request.messages`
  *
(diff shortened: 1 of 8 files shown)


nsdeschenes (approved):

Lms24 (approved):
good new test, thanks! FYI @alexander-alderman-webb for gen_ai attributes

PR #393, opened 2026-05-20
feat(schema): Add `additional_context` field to attribute definitions

matejminar:
Adds a new optional `additional_context` field to attribute definitions — a list of freeform notes capturing query-time nuances, common pitfalls, or behavioral context that isn't covered by `brief`.

## Motivation

Span attributes can have subtleties that trip up both users and AI agents when querying. For example, `gen_ai.cost.total_tokens` appears on both agent parent spans and LLM child spans, causing `sum()` queries to double-count. Today there's nowhere to capture this kind of context — `brief` describes *what* the attribute is, not *how to use it correctly*.

This is the foundation for surfacing attribute context to Explore UI (via Seer) and AI agent tooling. A follow-up PR will populate the field for gen_ai cost and usage attributes.

See proposal: https://www.notion.so/sentry/Proposal-Query-Guidance-for-Span-Attributes-3598b10e4b5d806d8c75e2a5b1caa42c

## Changes

- **Schema**: New optional `additional_context` property (array of strings) in `attribute.schema.json`
- **Types**: Added to `AttributeJson` interface in `types.ts`
- **Codegen**: Updated JS and Python generators to propagate the field into `AttributeMetadata` (as `additionalContext` in JS, `additional_context` in Python)
- **Generated code**: Regenerated JS and Python packages with the new field

Closes TET-2319, TET-2315

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -12605,6 +12605,8 @@ export interface AttributeMetadata {
   sdks?: string[];
   /** Changelog entries tracking how this attribute has changed across versions */
   changelog?: ChangelogEntry[];
+  /** A list of freeform notes providing additional context about how this attribute behaves, common pitfalls, or query-time nuances */
+  additionalContext?: string[];
 }
 
 export const ATTRIBUTE_TYPE: Record<string, AttributeType> = {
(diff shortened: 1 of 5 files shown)


Lms24 (approved):
nice, LGTM! We most likely also want to show this field in our Astro docs but we can tackle that as a follow-up PR once we also add the field to some attributes.

PR #394, opened 2026-05-20
feat(attributes): Add visibility field to attributes

nsdeschenes:
## Description
This PR adds in a new field to attributes: `Visibility`. This field will be used to determine if we want to filter out certain attributes from API responses as they're only used internally.

## PR Checklist


Lms24 (commented):
LGTM, thanks for backfilling all entries! 
One request for conventions: can we add/adjust a unit test to check that attribute entries have the field? We should already have such tests so should be fairly straight forward.

Lms24 (approved):

PR #395, opened 2026-05-20
feat(docs): Add filtering capability

nsdeschenes:
## Description

This PR adds in some selects to filter out attributes based off of different key fields that appear on attributes. You are also able to mix and match these filters to narrow down attributes even further.

These changes were done up by Codex and myself, everything seems to be working a-okay, but I'm not the best at Svelte 🫣

### DIFF
--- a/docs/src/components/AttributeFilters.svelte
+++ b/docs/src/components/AttributeFilters.svelte
@@ -273,7 +273,7 @@ const resultLabel = $derived.by(() => {
 });
 </script>
 
-<div class="mb-6 p-4 bg-bg-secondary border border-border rounded-lg">
+<div class="mb-6 p-3 bg-bg-secondary border border-border rounded-lg">
   <div class="flex flex-wrap items-end gap-3">
     <label class="flex flex-col gap-1 min-w-[140px]">
       <span class="text-xs text-text-muted uppercase tracking-wider">PII</span>
--- a/docs/src/components/AttributeFilters.svelte
+++ b/docs/src/components/AttributeFilters.svelte
@@ -1,0 +1,332 @@
+<script lang="ts">
+import { onMount, tick } from 'svelte';
+
+interface WindowWithAttributeIndex {
+  attributeIndex?: AttributeIndexEntry[];
+}
+
+interface AttributeIndexEntry {
+  key: string;
+  brief: string;
+  type: string;
+  category: string;
+  url: string;
+  deprecated: boolean;
+  pii: 'true' | 'maybe' | 'false';
+  visibility: 'public' | 'internal';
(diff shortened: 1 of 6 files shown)


nsdeschenes:
@cursor review

nsdeschenes:
@cursor review

docs/src/components/AttributeFilters.svelte
```diff
@@ -0,0 +1,350 @@
...
+  url: string;
+  deprecated: boolean;
+  pii: 'true' | 'maybe' | 'false';
+  visibility: 'public' | 'internal';
+  is_in_otel: boolean;
+}
+
+interface Props {
+  mode: 'cards' | 'sections';
+  category?: string;
```
Lms24: I think this prop isn't used, right?

docs/src/components/AttributeFilters.svelte
```diff
@@ -0,0 +1,350 @@
+<script lang="ts">
+import { onMount, tick } from 'svelte';
+
+interface WindowWithAttributeIndex {
+  attributeIndex?: AttributeIndexEntry[];
+}
```
Lms24: l: this seems to be unused

Lms24 (commented):
Nice! I gave the PR a try and filtering seems to work quite well. I think this is a pretty interesting approach because we render the entire list of attributes to static HTML and manually hide the ones not matching the filter criterion. A consequence of Astro being static-first and the filter component itself being the dynamic island. I think this is reasonable though. If we ever run into performance problems, we can convert the entire attributes list to be rendered as a svelte component but I don't think it's necessary to change this now. The Svelte component itself looks good to me!

Had a couple of nits and one visual request: 

Image

can we show "x/N attributes" instead of "N Attributes" on these labels?

nsdeschenes:
> Nice! I gave the PR a try and filtering seems to work quite well. I think this is a pretty interesting approach because we render the entire list of attributes to static HTML and manually hide the ones not matching the filter criterion. A consequence of Astro being static-first and the filter component itself being the dynamic island. I think this is reasonable though. If we ever run into performance problems, we can convert the entire attributes list to be rendered as a svelte component but I don't think it's necessary to change this now. The Svelte component itself looks good to me!
> 
> Had a couple of nits and one visual request:
> 
> Image
> can we show "x/N attributes" instead of "N Attributes" on these labels?

All feedback addressed, and added in that change for categories, though it will only show when we're rending the entire list. If a user has a selected category we leave it blank.

Lms24 (approved):
Nice, thanks for making the changes! I think the failing CI job comes from formatting. Might need to run `yarn format` again. 🚀

PR #400, opened 2026-05-21
fix: Change type of sentry.dsc.project_id to string

elramen:
## Description
Change the type of `sentry.dsc.project_id` from integer to string as there is no guarantee that a project id will fit in an i64 (u64 isn't supported by EAP). This prevents issues if we were to change the project id generation logic in the future.

Fixes [INGEST-927](https://linear.app/getsentry/issue/INGEST-927/update-sentrydscproject-id-in-sentry-conventions)

## PR Checklist


Lms24 (approved):
thanks!

Dav1dde (approved):

PR #401, opened 2026-05-21
feat(replays): Define `replayId` as a deprecated attribute

mjq:
## Description
SDKs currently send `replayId` on spans, but we'd like to standardize on `sentry.replay_id` (already defined and in use for other telemetry types). Marking it as deprecated lets the product transparently query either name and return results for either attribute.

## PR Checklist

### DIFF
--- a/model/attributes/replayId.json
+++ b/model/attributes/replayId.json
@@ -1,5 +1,5 @@
 {
-  "key": "sentry.replay_id",
+  "key": "replayId",
   "brief": "The id of the sentry replay.",
   "type": "string",
   "pii": {
@@ -7,11 +7,15 @@
   },
   "is_in_otel": false,
   "example": "123e4567e89b12d3a456426614174000",
-  "alias": ["replay_id", "replayId"],
+  "alias": ["sentry.replay_id"],
+  "deprecation": {
+    "_status": null,
+    "replacement": "sentry.replay_id"
+  },
   "visibility": "public",
   "changelog": [
     {
-      "version": "0.0.0"
+      "version": "next"
     }
   ]
 }
--- a/model/attributes/replayId.json
+++ b/model/attributes/replayId.json
@@ -1,5 +1,5 @@
 {
-  "key": "sentry.replay_id",
+  "key": "replayId",
   "brief": "The id of the sentry replay.",
   "type": "string",
   "apply_scrubbing": {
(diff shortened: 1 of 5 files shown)


model/attributes/replayId.json
```diff
@@ -0,0 +1,21 @@
...
+  "example": "123e4567e89b12d3a456426614174000",
+  "alias": ["sentry.replay_id"],
+  "deprecation": {
+    "_status": null,
+    "replacement": "sentry.replay_id"
+  },
+  "visibility": "public",
+  "changelog": [
+    {
+      "version": "next"
```
Lms24, line 18: nit: we can link to this PR for reference

```suggestion
      "version": "next"
      "prs": [401]
```

Lms24 (approved):

model/attributes/replayId.json
mjq: Double-checking - is this necessary? I used to think it was, but it looks like it's actually handled automatically by the generation script (the one that rewrites `next` into the version number).

https://github.com/getsentry/sentry-conventions/blob/6ff072db52bce1041b276dc60cc6ade44476b9e3/scripts/generate_attribute_changelog.ts#L289-L292

https://github.com/getsentry/sentry-conventions/blob/6ff072db52bce1041b276dc60cc6ade44476b9e3/scripts/generate_attribute_changelog.ts#L116-L140

Lms24: The `generate_attribute_changelog.ts` script currently isn't called at release time. AFAIK, this was meant to run manually once but not repeatedly (will check and delete if so). For releases, we only the `bump_attribute_changelog.ts` script. But this one only does the `"next"` to `{new_version}` conversion but doesn't add the PR number. Something we should fix! I'll look into this. 

For this PR, no strong feelings. Feel free to merge with or without the PR reference!

Lms24: I remembered why `generate_attribute_changelog` isn't called at release: It takes 5+ minutes locally to run because it looks at every attribute x every single version of conventions to "backfill" PRs and version data from git history. I think we can make this a bit faster for releases by limiting it to "next" entries.

mjq: Ahhh, thanks for the context! Makes sense. I'll add the PR ID before taking this out of draft. (Need product changes before I can safely merge this).

nsdeschenes (approved):
LGTM!

PR #407, opened 2026-05-30
feat(attributes): Add db.operation.batch.size

isaacs:
## Description

Add the OTel span attribute [`db.operation.batch.size`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-operation-batch-size)

## PR Checklist

If an attribute was added:
- [ x I have used the correct value for `pii` (i.e. `maybe` or `true`. Use `false` only for values that should never be scrubbed such as IDs)

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -3330,6 +3330,27 @@ export const DB_OPERATION = 'db.operation';
  */
 export type DB_OPERATION_TYPE = string;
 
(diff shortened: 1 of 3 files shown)


Lms24 (approved):
Once https://github.com/getsentry/sentry-conventions/pull/408 is merged and this branch is rebased, CI should pass. Changes LGTM!

(given this is an OTel attribute, feel free to skip the 3 day wait period and merge sooner)

PR #408, opened 2026-06-01
fix(cloudfare): add missing visibility attributes

markushi:
## Description

Looks like https://github.com/getsentry/sentry-conventions/pull/392 introduced new props without the visibility flag, so adding "public" to all of them.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -2338,6 +2338,7 @@ export type CLOUDFLARE_D1_DURATION_TYPE = number;
  * Contains PII: false
  *
  * Attribute defined in OTEL: No
+ * Visibility: public
  *
  * @example "run"
  */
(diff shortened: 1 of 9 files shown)


JPeer264 (approved):
Interesting. I actually used the `yarn` command from the README, maybe that should be added to the script?

markushi:
> Interesting. I actually used the `yarn` command from the README, maybe that should be added to the script?

I think it was just a timing issue, the PR was approved before the visibility flag was introduced.

Lms24 (approved):
thanks!

cleptric:
Should we add some linter for this?

Lms24:
We have tests that fail for this. On new PRs, these tests would fail whenever a lint job fails. I think the issue here was just that we would have needed to rebase PRs prior to merging. Should we enforce that the branch must be up to date prior to merging?

Lms24:
Gonna go ahead and merge this to unblock other PRs

PR #413, opened 2026-06-03
feat(attributes): Add Cloudflare R2 attributes

JPeer264:
## Description

This adds Cloudflare R2 attributes. These attributes were taken from Cloudflare's traces directly.

Only `cloudflare.r2.bucket` is not part of Cloudflare itself, and was inspired by [AWS S3](https://opentelemetry.io/docs/specs/semconv/object-stores/s3/) attributes.

For the PII I wasn't sure if that is correct, maybe one could have a special eye on these values.

## PR Checklist

If an attribute was added:

### DIFF
--- a/model/attributes/cloudflare/cloudflare__r2__bucket.json
+++ b/model/attributes/cloudflare/cloudflare__r2__bucket.json
@@ -1,0 +1,18 @@
+{
+  "key": "cloudflare.r2.bucket",
+  "brief": "The name of the Cloudflare R2 bucket binding",
+  "type": "string",
+  "pii": {
+    "key": "false"
+  },
+  "is_in_otel": false,
+  "visibility": "public",
+  "example": "MY_BUCKET",
+  "changelog": [
+    {
+      "version": "next",
+      "prs": [413],
+      "description": "Added cloudflare.r2.bucket attribute"
+    }
+  ]
+}
--- a/model/attributes/cloudflare/cloudflare__r2__bucket.json
+++ b/model/attributes/cloudflare/cloudflare__r2__bucket.json
@@ -3,7 +3,7 @@
   "brief": "The name of the Cloudflare R2 bucket binding",
   "type": "string",
   "pii": {
-    "key": "false"
+    "key": "maybe"
   },
   "is_in_otel": false,
   "visibility": "public",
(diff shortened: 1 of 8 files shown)


cleptric:
Set pii to `maybe`. With `false`, users can't setup scrubbing rules.

model/attributes/cloudflare/cloudflare__r2__request__part_number.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "cloudflare.r2.request.part_number",
+  "brief": "The part number in a multipart upload operation",
+  "type": "integer",
+  "pii": {
+    "key": "false"
```
Lms24: l: let's go with `'maybe'` here. This only means that users _can_ configure scrubbing for this field (while they can't for `false`). IIRC we said that all number fields should get `'maybe'`

```suggestion
    "key": "maybe"
```

model/attributes/cloudflare/cloudflare__r2__bucket.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "cloudflare.r2.bucket",
+  "brief": "The name of the Cloudflare R2 bucket binding",
+  "type": "string",
+  "pii": {
+    "key": "false"
```
Lms24:
```suggestion
    "key": "maybe"
```

model/attributes/cloudflare/cloudflare__r2__operation.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "cloudflare.r2.operation",
+  "brief": "The R2 API operation being performed",
+  "type": "string",
+  "pii": {
+    "key": "false"
```
Lms24:
```suggestion
    "key": "maybe"
```

model/attributes/cloudflare/cloudflare__r2__request__delimiter.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "cloudflare.r2.request.delimiter",
+  "brief": "The delimiter used to group objects in an R2 list operation",
+  "type": "string",
+  "pii": {
+    "key": "false"
```
Lms24:
```suggestion
    "key": "maybe"
```

Lms24:
(ah sorry, didn't see Michi's review)

JPeer264:
Moved all to `maybe`

Lms24 (approved):

PR #416, opened 2026-06-08
feat(attributes): Add `host.id` attribute

alexander-alderman-webb:
## Description

Document the field currently set as context by `sentry-python`:

https://github.com/getsentry/sentry-python/blob/7d38a49af058d46299e1e01ec0086ff05f585866/sentry_sdk/integrations/cloud_resource_context.py#L130

The host ID would be set as an attribute with the streaming trace lifecycle.

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -6540,7 +6540,7 @@ export type HARDWARECONCURRENCY_TYPE = string;
 // Path: model/attributes/host/host__id.json
 
 /**
- * Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. For non-containerized systems, this should be the `machine-id`. `host.id`
+ * Unique host ID. In cloud environments, this must be the instance ID assigned by the cloud provider. `host.id`
  *
  * Attribute Value Type: `string` {@link HOST_ID_TYPE}
  *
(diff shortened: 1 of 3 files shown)


javascript/sentry-conventions/src/attributes.ts
```diff
@@ -6537,6 +6537,27 @@ export const HARDWARECONCURRENCY = 'hardwareConcurrency';
  */
 export type HARDWARECONCURRENCY_TYPE = string;
 
+// Path: model/attributes/host/host__id.json
+
+/**
+ * Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. For non-containerized systems, this should be the `machine-id`. `host.id`
```
cleptric: For Cloud is weird english 😬

alexander-alderman-webb: Agreed, this is copied from the OTel description

https://opentelemetry.io/docs/specs/semconv/resource/host/

cleptric: We can change things.

alexander-alderman-webb: Perfect, I've updated the description to be how I would phrase it.

Lms24 (approved):

PR #419, opened 2026-06-09
meta: Add note about updating conventions in Relay

loewenheim:


tobias-wilfert (approved):

Lms24 (approved):
Thanks for clarifying, I always wondered how this works!

side-note: Sounds like we could add some automation to always trigger this on a conventions release 😅 Though not sure about your preferred cadence of updating conventions in Relay. Does it make sense to couple them to convention releases? Or do we want them more/less often?

buenaflor (approved):
We should do the same for Sentry and Snuba

PR #421, opened 2026-06-09
feat(attributes): Add `process.runtime.build` attribute

alexander-alderman-webb:
## Description

Document the field currently set as context by sentry-python:

https://github.com/getsentry/sentry-python/blob/3bbd78d4fe3a2803839d8091ecc435136bd83363/sentry_sdk/integrations/stdlib.py#L39

The string would be set as an attribute with the streaming trace lifecycle.

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -10195,6 +10195,29 @@ export const PROCESS_PID = 'process.pid';
  */
 export type PROCESS_PID_TYPE = number;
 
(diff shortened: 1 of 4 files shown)


ericapisani:
I don't think we need this attribute - there's a `process.runtime.description` OTEL property that we already have within our conventions that [OTEL says we should use for `sys.version`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/process.md#python-runtimes)

alexander-alderman-webb:
Nice don't need this then!

PR #424, opened 2026-06-11
ref(pii): Review pii settings for attributes

Dav1dde:
- `pii=false` should only be used in the `sentry` namespace and even then, it's very rare that it is needed
- Changes all non sentry.* attributes from `pii=false` to a less strict variant (true/maybe)
- Changes a bunch of attributes which can contain pii to `true` (http body, urls, parameters, etc.)
- Some rare fields which do not contain user information but infra information are changed from `true` to `maybe`.

### DIFF
--- a/model/attributes/browser/browser__web_vital__lcp__url.json
+++ b/model/attributes/browser/browser__web_vital__lcp__url.json
@@ -3,7 +3,7 @@
   "brief": "The url of the dom element responsible for the largest contentful paint",
   "type": "string",
   "pii": {
-    "key": "maybe"
+    "key": "true"
   },
   "is_in_otel": false,
   "example": "https://example.com/static/img.png",
--- a/model/attributes/db/db__statement.json
+++ b/model/attributes/db/db__statement.json
@@ -3,7 +3,7 @@
   "brief": "The database statement being executed.",
   "type": "string",
   "pii": {
-    "key": "maybe"
+    "key": "true"
   },
   "is_in_otel": true,
   "example": "SELECT * FROM users",
--- a/model/attributes/device/device__id.json
+++ b/model/attributes/device/device__id.json
@@ -3,7 +3,7 @@
   "brief": "Unique device identifier.",
   "type": "string",
   "pii": {
-    "key": "maybe"
+    "key": "true"
   },
   "is_in_otel": true,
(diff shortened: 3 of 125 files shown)


cleptric (approved):

model/attributes/browser/browser__web_vital__lcp__url.json
```diff
@@ -3,7 +3,7 @@
   "brief": "The url of the dom element responsible for the largest contentful paint",
   "type": "string",
   "pii": {
-    "key": "maybe"
+    "key": "true"
```
jjbayer, line 6: DOM elements are usually public, does it really make sense to scrub them by default?

loewenheim (approved):

model/attributes/db/db__statement.json
```diff
@@ -3,7 +3,7 @@
   "brief": "The database statement being executed.",
   "type": "string",
   "pii": {
-    "key": "maybe"
+    "key": "true"
```
jjbayer, line 6: The changes to db attributes make sense to me but could definitely have product impact, cc @mjq.

model/attributes/device/device__id.json
```diff
@@ -3,7 +3,7 @@
   "brief": "Unique device identifier.",
   "type": "string",
   "pii": {
-    "key": "maybe"
+    "key": "true"
```
jjbayer, line 6: What default scrubbing rules do we expect to match on a device ID? This seems more like a `maybe` to me.

jjbayer (commented):
Changes from `false -> maybe` definitely make sense. Changes from `maybe` to `true` might negatively affect the UX (famous example is the `auth` matcher replacing words like `Author`).

model/attributes/browser/browser__web_vital__lcp__url.json
Dav1dde: This is the URL, right? Might be `http://domain.host/foo?userId=123`

model/attributes/device/device__id.json
Dav1dde: damn it, I fixed that but never pushed yesterday
image

Dav1dde: Addressed in #425

PR #431, opened 2026-06-16
build: Bump all dependencies to fix esbuild security issues

cleptric:
_This was done purely with Claude Code_

---

Bump all direct dependencies to their latest versions and add a `resolutions` override to force `esbuild >=0.28.1` across all transitive consumers (astro, vite, unbuild, mkdist). This addresses two high-severity esbuild advisories: GHSA-gv7w-rqvm-qjhr (RCE via `NPM_CONFIG_REGISTRY` in Deno) and GHSA-g7r4-m6w7-qqqr (arbitrary file read on Windows dev server).

**esbuild Resolution Override**

Upstream packages (astro, vite, unbuild, mkdist) declare `^0.25.x`/`^0.27.x` ranges for esbuild, which in the 0.x semver space won't resolve past their minor version. A root-level `resolutions` entry forces all instances to the patched 0.28.1.

**Biome 2.5.0 Migration**

The `@biomejs/biome` bump from 1.9.4 to 2.5.0 required a config schema migration: `ignore` → `includes`, `organizeImports` → `assist`, `recommended` → `preset: "recommended"`, and enabling `css.parser.tailwindDirectives` for the docs site's Tailwind CSS.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,9 +1,17 @@
 {
-  "$schema": "https://biomejs.dev/schemas/1.9.3/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.5.0/schema.json",
   "vcs": { "enabled": false, "clientKind": "git", "useIgnoreFile": false },
   "files": {
     "ignoreUnknown": false,
-    "ignore": ["dist/*", "javascript/sentry-conventions/package.json", "rust/*", "python/*", ".claude/*", ".astro/*"]
+    "includes": [
+      "**",
+      "!**/dist/**/*",
+      "!**/javascript/sentry-conventions/package.json",
+      "!**/rust/**/*",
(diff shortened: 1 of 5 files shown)


cleptric:
@Lms24 can you check this please? Maybe there is a more elegant solution for esbuild. Also we might want to migrate this all to the void(zero) stack.

Lms24 (approved):
Thanks for updating! Seems like the resolutions approach is our best bet for now until the underlying packages bump to the safe esbuild version (we need the resolution despite the packages declaring a `^` dependency on esbuild because esbuild is on `0.x`, letting yarn not automatically resolve the dep to the latest minor version). I changed the resolution version range from `>=` to `~` just to be on the safe side.

PR #433, opened 2026-06-17
feat(docs): Make span name and description cards linkable

Lms24:
Adds anchor links to span name and description cards so that we can link to specific name/description rules more easily.


mjq (approved):
🔗🎉

PR #435, opened 2026-06-18
feat(attributes): Add Cloudflare DO sql attributes

JPeer264:
## Description

This adds Cloudflare attributes for SQL in Durable Object (sqlite). These attributes were taken from Cloudflare's traces directly.

## PR Checklist

If an attribute was added:


cleptric (changes requested):
As this is just SQL, I would say we stick to the generic `db` ones. Or does Cloudflare's own instrumentation uses those?

JPeer264:
> Or does Cloudflare's own instrumentation uses those?

Yes Cloudflare uses exactly these. `db` is of course also there, so these will be used anyways. The following is a snapshot from a `sql.exec` method. In theory we can also just don't emit these extra attributes:

image

cleptric (approved):

Lms24 (approved):

PR #439, opened 2026-06-19
docs: Remove semicolon in db.query.summary example

JPeer264:
## Description

I was going over the `db.query.*` examples and saw that the `;` should not be there in `db.query.summary` this also reflects the conventions: https://opentelemetry.io/docs/specs/semconv/db/sql/ (it is just a little tough to see, as there is a `;`, but it is not part of the example)

## PR Checklist

If an attribute was added:


Lms24 (approved):
Good catch, thanks! Honestly, I had to look 3 times at the OTel docs to grasp that the semicolon isn't part of the examples, even after you called it out lol

PR #442, opened 2026-06-19
feat(attributes): Add `sentry.trace_lifecycle` attribute

Lms24:
## Description

This PR proposes adding a `sentry.trace_lifecycle` attribute to spans that indicates if a span was sent by the SDK as a streamed span (or not, i.e. as a static (transaction-sent) span). It's quite difficult at the moment to indicate in the product that a span is actually a streamed span. This allows us to explicitly query for this attribute. It also helps us analyze how many users will switch/opt-out of span streaming once it becomes the default in the next SDK major.

## PR Checklist

If an attribute was added:


cleptric (approved):
Good addition.

buenaflor (approved):

PR #443, opened 2026-06-19
feat(attributes): Add `span.initial_name` attribute

Lms24:
## Description

This PR proposes adding a new attribute set on every span that carries the initial span name. We can surface this attribute in the product to make it easier for users to see what they have to put into `ignoreSpans` to filter out spans they don't want. In contrast to the span name, this attribute will always hold the original name and not get updated later on by auto instrumentation or users. 

cc @ArthurKnaus 

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -12695,6 +12695,27 @@ export const SESSION_ID = 'session.id';
  */
 export type SESSION_ID_TYPE = string;
 
+// Path: model/attributes/span/span__initial_name.json
+
+/**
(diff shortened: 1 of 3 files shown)


alexander-alderman-webb (approved):

Lms24:
The counter argument to merging this: We've been thinking of adding more metadata to telemetry properties to track changes to these over time (e.g. SDK updates span name, Relay scrubs or adds attribute, something in the pipeline truncates an attribute, etc). I'm also happy to close this proposal and wait for a more well-thought out experience

Lms24:
Gonna close for the moment due to uncertainty regarding attribute metadata/processing annotations. Happy to reopen if we need this urgently

PR #445, opened 2026-06-22
feat(name): Add resource span name model

cleptric:
Define span naming conventions for browser resource loading operations. Uses `url.domain` as the primary template with a static `Resource` fallback when the attribute is unavailable.

Covers all resource ops: `resource`, `resource.script`, `resource.css`, `resource.link`, `resource.img`, `resource.audio`, `resource.video`, `resource.iframe`, and `resource.other`.


Lms24 (approved):

model/name/resource.json
```diff
@@ -0,0 +1,23 @@
...
+        "resource.css",
+        "resource.link",
+        "resource.img",
+        "resource.audio",
+        "resource.video",
+        "resource.iframe",
+        "resource.other"
+      ],
+      "templates": ["{{url.domain}}", "Resource"],
+      "examples": ["https://example.com/static/app.js"]
```
Lms24:
```suggestion
      "examples": ["example.com"]
```

loewenheim (approved):

PR #452, opened 2026-06-24
feat(docs): Add span operations reference page

Lms24:
Add a Span Operations page displaying the span operations data (`op` schema). 

There are some inconsistencies here like missing ops and op descriptions, but this is rather a data issue to be addressed in a follow-up. Ideally we can phase out the develop docs [span operations](https://develop.sentry.dev/sdk/telemetry/traces/span-operations/#list-of-operations) list since in conventions we can much easier cross-check and reference ops and their influence on span name/description rules.

Similarly, as a follow-up I'm thinking of adding tests that for span description inference rules, the ops must be present in the operations page. At this point we can also link from description and name rule ops to the respective op entry. 

image

### DIFF
--- a/docs/src/utils/markdown.ts
+++ b/docs/src/utils/markdown.ts
@@ -1,0 +1,31 @@
+function escapeHtml(value: string): string {
+  return value
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+}
+
+/**
+ * Renders a string containing inline markdown links (`[label](https://...)`) to HTML.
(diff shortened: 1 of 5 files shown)


docs/src/utils/markdown.ts
```diff
@@ -0,0 +1,31 @@
...
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+}
+
+/**
+ * Renders a string containing inline markdown links (`[label](https://...)`) to HTML.
+ * Everything except recognized http(s) links is HTML-escaped, so the output is safe to
+ * pass to `set:html`.
+ */
+export function renderMarkdownLinks(text: string): string {
```
Lms24, line 15: a small helper since we had some markdown links in the op descriptions. Might be neat to use this on other pages as well but keeping it local for the moment

nsdeschenes (approved):

PR #454, opened 2026-06-26
feat(attributes): add data migration utilities

constantinius:
## Summary

- add shared data migration APIs for deprecated attributes across JS, Python, and Rust
- expose migration metadata in generated attribute metadata
- use open-ended string migration identifiers instead of enumerating ids in schemas or generated types
- document how SDKs and Relay should use the migration helpers and how contributors add future data migrations

Contributes to TET-2369.

## Testing

- `yarn test`


loewenheim:
Hi, we're going to discuss in the ingest team how this API would fit into Relay. In the meantime, a question: is the migration machinery intended to eventually replace the existing backfill/normalize logic?

constantinius:
Thank you

> is the migration machinery intended to eventually replace the existing backfill/normalize logic?

No, I would say it complements it. The migrations are only used when we change the data layout. But we should discuss how it works together: e.g backfill + migration keeps the old value on the old field, and sets the new value on the new field.

loewenheim (changes requested):
This needs some further offline discussion.

alexander-alderman-webb:
My knee-jerk reaction to this: why don't we design schemas that are forwards compatible instead?

constantinius:
Closing in favor of #465, which replaces the migration-based approach with attribute transformation documents and transform deprecation references.

PR #455, opened 2026-06-26
feat(attributes): add gen ai data migrations

constantinius:
## Summary

- add `gen_ai.request.messages` → `gen_ai.input.messages` data migration
- add `gen_ai.response.text` / `gen_ai.response.tool_calls` → `gen_ai.output.messages` data migration
- annotate source and target attributes with string migration identifiers
- cover the migrations across JS, Python, and Rust

Stacked on #454.

Contributes to TET-2369.

## Testing

- `yarn test`


constantinius:
Closing in favor of #465, which replaces the migration-based approach with attribute transformation documents and transform deprecation references.

PR #457, opened 2026-06-30
fix(build): Do not reference const in object

JPeer264:
## Description

The only thing which changes is the way how the constants are used within an object:

```diff
export const ATTRIBUTE_TYPE: Record<string, AttributeType> = {
-  [AI_CITATIONS]: 'string[]',
+  'ai.citations': 'string[]',
};
```

When having following code ESBuild is not able to tree-shake `ATTRIBUTE_TYPE`:

```js
import { ANY_CONSTANT } from '@sentry/conventions';

console.log(ANY_CONSTANT);
```

When removing the references in the objects ESBuild is now able to tree-shake `ATTRIBUTE_TYPE`, as the constants are not referenced anymore 🤷 

seems like this is the reason: https://github.com/evanw/esbuild/issues/4373

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -14760,674 +14760,674 @@ export interface AttributeMetadata {
 }
 
 export const ATTRIBUTE_TYPE: Record<string, AttributeType> = {
-  [AI_CITATIONS]: 'string[]',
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
Thanks for looking into and fixing this!

(I'm gonna cut a conventions release after we merge this so that we can properly use the package in the JS SDK)

PR #458, opened 2026-06-30
feat(attributes): Add `file.path` and `file.size`

buenaflor:
Add the OTel-aligned `file.path` and `file.size` attributes to the `file` namespace.

Both attributes are already defined in the [OpenTelemetry semantic conventions registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/), so per the contributing guidelines we reuse the OTel names and types rather than introducing Sentry-specific synonyms, and set `is_in_otel: true`.

- `file.path` — `string`, e.g. `/home/user/example.txt`
- `file.size` — `integer` (bytes), e.g. `1024`


mjq (approved):

model/attributes/file/file__path.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "file.path",
+  "brief": "Path to the file.",
+  "type": "string",
+  "apply_scrubbing": {
+    "key": "manual"
```
Lms24: m: should we raise this to `"auto"`? I imagine this attribute could contain end user paths (e.g. from stack traces). not PII but could be considered sensitive. Feel free to disregard

buenaflor: Switched to `auto` in dc0dc54. Makes sense imo

Lms24 (approved):

PR #460, opened 2026-07-02
feat(attributes): expand grpc attributes

lucas-zimerman:
## Description

Adds attribute conventions for the Google gRPC rich error model (`google.rpc.error_details.proto`), under the `grpc.error.*` namespace:

- `grpc.error.error_info.{reason,domain,metadata.<key>}`
- `grpc.error.bad_request.field_violations`
- `grpc.error.retry_info.retry_delay_ms`
- `grpc.error.debug_info.{detail,stack_entries}`
- `grpc.error.precondition_failure.violations`
- `grpc.error.resource_info.{resource_type,resource_name,description,owner}`
- `grpc.error.quota_failure.violations`

Field names and shapes were verified against `google/rpc/error_details.proto` (AIP-193 error model) for fidelity. Structured/repeated fields (e.g. `field_violations`, `violations`) are encoded as `JSON-stringified` array entries, following the existing `mcp.request.argument.<key>` precedent. Attributes that may carry user- or resource-identifying data (e.g. `error_info.metadata.<key>`, `debug_info.detail`, `debug_info.stack_entries`, `precondition_failure.violations`, `resource_info.resource_name`, `resource_info.owner`, `quota_failure.violations`) are documented as PII and gated behind `sendDefaultPii` or Sentry's `dataCollection` configuration.

Unblocks getsentry/sentry-dart#3791.

## PR Checklist

- [x] I have run yarn test and verified that the tests pass.
- [x] I have run yarn generate to generate and format code and docs.

If an attribute was added:

- [x] The attribute is in a namespace (e.g. nextjs.function_id, not function_id)
- [x] I have used the correct value for apply_scrubbing (i.e. manual or auto. Use never only for values that should never be scrubbed such as IDs)


lucas-zimerman:
I used claude to fix the `attributes.py`, not sure why when I generate it manually it formats the entire file
image

buenaflor (commented):
I'm not sure if we should nest under rpc, gRPC themselves use a top level `grpc.*` (at least for metrics)

https://grpc.io/docs/guides/opentelemetry-metrics/

model/attributes/rpc/rpc__grpc__error__debug_info__detail.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "rpc.grpc.error.debug_info.detail",
+  "brief": "Additional debugging information, such as a server-side stack trace, from a google.rpc.DebugInfo error detail. SDKs should only send this attribute when sendDefaultPii is enabled.",
```
Lms24: Since we're phasing out `sendDefaultPii`, let's mention [dataCollection](https://develop.sentry.dev/sdk/foundations/client/data-collection/) as well. This applies to all attributes where we meniton sendDefaultPii at the moment.

```suggestion
  "brief": "Additional debugging information, such as a server-side stack trace, from a google.rpc.DebugInfo error detail. SDKs should only send this attribute when sendDefaultPii is enabled or dataCollection is configured accordingly.",
```

Lms24 (commented):
Theoretically, we have the `error` namespace (also [in OTel](https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/)) which _"defines the shared attributes used to report an error"_. But I get that the attributes here are very specific to GRPC, with some exceptions. 

Also to confirm: We don't capture an exception in the SDK because we explicitly only want to record the error on the span? I assume because the error is not from the application but happened wherever the procedure was executed?

I'm curious on other reviewers opinions here. Should we stay GRPC-specific or would we prefer adding a bunch of specific attributes to the `error` namespace?

lucas-zimerman:
> I'm not sure if we should nest under rpc, gRPC themselves use a top level `grpc.*` (at least for metrics)
> 
> https://grpc.io/docs/guides/opentelemetry-metrics/

renamed to `grpc.*` on https://github.com/getsentry/sentry-conventions/pull/460/commits/f0c8eed367808bc63eeec96e437d8f9a9f0f9bc1

Except for `rpc.grpc.status_code` that already existed on sentry-conventions.

lucas-zimerman:
> Theoretically, we have the `error` namespace (also [in OTel](https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/)) which _"defines the shared attributes used to report an error"_. But I get that the attributes here are very specific to GRPC, with some exceptions.
> 
> Also to confirm: We don't capture an exception in the SDK because we explicitly only want to record the error on the span? I assume because the error is not from the application but happened wherever the procedure was executed?
> 
> I'm curious on other reviewers opinions here. Should we stay GRPC-specific or would we prefer adding a bunch of specific attributes to the `error` namespace?

OTel's error.* namespace is intentionally minimal: it's really just error.type (a low-cardinality classifier like an exception class name or domain code), plus a deprecated error.message. It's explicitly not meant to carry rich, structured payloads. The docs even call out that message-like fields cause "unbounded cardinality" problems on spans.


Also. Capturing an exception client-side felt low-value here. The client only has the serialized `google.rpc` error detail, while the real exception with full context typically already exists server-side (when that service is also Sentry-instrumented). Turning every non-OK status into a client-side exception would also add noise for cases that are expected/handled control flow rather than bugs. Worth documenting that reasoning, but happy to make it opt-in if reviewers want richer client-side capture for `unattended`/`uninstrumented` peers.

model/attributes/grpc/grpc__error__bad_request__field_violations.json
```diff
@@ -0,0 +1,20 @@
...
+  "apply_scrubbing": {
+    "key": "manual"
+  },
+  "is_in_otel": false,
+  "visibility": "public",
+  "example": [
+    "{\"field\":\"email\",\"description\":\"must be a valid email address\",\"reason\":\"FIELD_INVALID\",\"localized_message\":{\"locale\":\"en-US\",\"message\":\"Must be a valid email address\"}}"
+  ],
+  "changelog": [
+    {
```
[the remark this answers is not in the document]
lucas-zimerman: My understanding is `manual` could be ok? `never` would be the problem here

model/attributes/grpc/grpc__error__retry_info__retry_delay_in_ms.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "grpc.error.retry_info.retry_delay_in_ms",
```
Lms24: l: OTel [encourages](https://opentelemetry.io/docs/specs/semconv/general/naming/#:~:text=Use%20shorter%20names%20when%20it%20does%20not%20compromise%20clarity) shorter names. I think we can omit the `_in` here but LOGAF-L.

```suggestion
  "key": "grpc.error.retry_info.retry_delay_ms",
```

Lms24 (commented):
> I used claude to fix the attributes.py, not sure why when I generate it manually it formats the entire file

You need to use `yarn generate` to generate the file. It shouldn't reformat the entire file though. We recently switched to `oxfmt`. Maybe you need to run `yarn install` again (?)

>> I'm not sure if we should nest under rpc, gRPC themselves use a top level grpc.* (at least for metrics)

> renamed to grpc.* on https://github.com/getsentry/sentry-conventions/commit/f0c8eed367808bc63eeec96e437d8f9a9f0f9bc1

So I researched this and I agree, we should just introduce the `grpc` namespace. For some context, I'm gonna write down what I learned here. The reason for going to the dedicated `grpc` namespace is that OTel [deprecated](https://opentelemetry.io/docs/specs/semconv/registry/attributes/rpc/#rpc-grpc-status-code) a bunch of `rpc.grpc.*` attributes. Ideally they should be replaced with top-level `rpc` attributes. But in a case like gRPC rich errors, this isn't really applicable, since it's a gRPC-specific concept. In this case, the [T-shaped signals](https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/t-shaped-signals/) spec recommends creating a new specialized attribute. There's also precedence for this with `jsonrpc`, where they already [moved](https://opentelemetry.io/docs/specs/semconv/registry/attributes/rpc/#rpc-jsonrpc-request-id) from `rpc.jsonrpc.*` to `jsonrpc.*`. So it seems like OTel just didn't (yet) see a need for a `grpc` namespace, but we can do so :) 

 > Except for rpc.grpc.status_code that already existed on sentry-conventions.

We should deprecate this field in favour of `rpc.response.status_code` [like OTel did](https://opentelemetry.io/docs/specs/semconv/registry/attributes/rpc/#rpc-grpc-status-code). I'd appreciate it if you could take care of this, but please in a follow-up PR, since it's not directly related to this one.

Lms24 (approved):
Gonna approve since the remaining changes are minimal to kick off the 3-day period:

- please update the PR description with the final attribute names
- look at my comment regarding `_ms` vs. `_in_ms` 
- would appreciate the follow-up PR with the status code deprecation (see comment above)

model/attributes/grpc/grpc__error__retry_info__retry_delay_in_ms.json
lucas-zimerman: Resolved on https://github.com/getsentry/sentry-conventions/pull/460/commits/d86e8ca0634be66b44b750e9cf6bee2b18e629c4

lucas-zimerman:
> Gonna approve since the remaining changes are minimal to kick off the 3-day period:
> 
>     * please update the PR description with the final attribute names
> 
>     * look at my comment regarding `_ms` vs. `_in_ms`
> 
>     * would appreciate the follow-up PR with the status code deprecation (see comment above)

PR description updated, also updated the `_ms` field. Happy to make a follow-up PR too.

Also, I can't merge this PR since I am an external contributor here

buenaflor (approved):

buenaflor:
CI is failing cc @lucas-zimerman

Lms24:
I think these are some weird installation flakes. Already raised this internally. for now, we need to rerun CI until installations pass

Lms24:
oh actually, sorry, in this case, it's a test fail. we have to regenerate code due to #471 

(`yarn generate`)

lucas-zimerman:
I ran the same test locally and it seems to be passing now

javascript/sentry-conventions/src/attributes.ts
```diff
@@ -7176,6 +7176,281 @@ export const GRAPHQL_OPERATION_TYPE = 'graphql.operation.type';
...
+ * Has Dynamic Suffix: true
+ *
+ * @example "grpc.error.error_info.metadata.user_id='123'"
+ */
+export const GRPC_ERROR_ERROR_INFO_METADATA_KEY = 'grpc.error.error_info.metadata.<key>';
+
+/**
+ * Type for {@link GRPC_ERROR_ERROR_INFO_METADATA_KEY} grpc.error.error_info.metadata.<key>
+ */
+export type GRPC_ERROR_ERROR_INFO_METADATA_KEY_TYPE = string;
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed on ded27a1

PR #465, opened 2026-07-07
feat(attributes): add attribute transformations

constantinius:
## Summary

- add attribute transformation documents for deprecated attributes whose replacement values need reshaping
- add `deprecation._status: "transform"` and `deprecation.transformation` references
- document how to add attribute transformations
- expose transformation references in generated JS and Python metadata

This replaces the previous migration-based approach from #454 and #455.

Contributes to TET-2369.
Closes TET-2586.

## Testing

- `yarn test`


alexander-alderman-webb:
The types set for `gen_ai.request.messages` and `gen_ai.response.text` are inconsistent in the SDKs.
You yourself have added the `litellm` integration that sets a list of objects for `gen_ai.response.text`.
How should the suggested transformations be implemented if the source attribute does not have a consistent type?

constantinius:
> How should the suggested transformations be implemented if the source attribute does not have a consistent type?

This is true and indeed an issue. In the frontend, we have several paths to get the input/output messages, I will match them in Relay, but I think I should also mention them here in the sentry-conventions

model/attribute_transformations/gen_ai_request_messages_to_input_messages.json
```diff
@@ -0,0 +1,45 @@
+{
+  "id": "gen_ai_request_messages_to_input_messages",
```
jjbayer, line 2: Is the ID also what we expect the function name to be in Relay, by convention?

constantinius: No, this is just the ID we link to from the attribute where we want to apply the transformation.

constantinius:
@alexander-alderman-webb 
This now also includes the transformation from various output formats we encounter across the SDK integrations.

alexander-alderman-webb:
Well, that's not true 😅.
For example, the LangChain integration sets an array of arrays as `gen_ai.response.text`.

constantinius:
@alexander-alderman-webb 

> For example, the LangChain integration sets an array of arrays as gen_ai.response.text.

I can't confirm this. In the testing framework I get for Python/JS LangChain only strings or array of strings across all tests. Do you have a working example?

Also, at some point we have to stop chasing wrong implementations.

loewenheim (approved):
LGTM. The extensive specification and examples are nice.

PR #472, opened 2026-07-09
feat(attribute): Add `koa.type` and (deprecated) `koa.name`

s1gr1d:
## Description

Adds `koa.type` which is used in the koa instrumentation: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/packages/instrumentation-koa/src/enums/AttributeNames.ts

`koa.name` can be substituted with either `http.route` (for routes) or `code.function.name` (for middleware) as middleware is named after the function name: https://github.com/koajs/koa/blob/master/docs/guide.md#named-middleware

## PR Checklist

If an attribute was added:


Lms24 (approved):
Thanks for adding! 

RE: `koa.name`: We can still add it to conventions and mark it as deprecated. Add `_status: null` to the deprecation since there's no direct replacement. In this case, the SDK will decide which attributes replace it in which case. The benefit is still that the attribute is documented and that it's gonna be deprecated (i.e. signaling conventions consumers that it shouldn't be set anymore). But feel free to tackle later

PR #476, opened 2026-07-09
feat(attributes): Add `sentry.metric.source`

klochek:
## Description

This is an attribute that can be used to attribute the source of a metric; for example, Relay can use it to indicate that a metric originated from an ingested span.

## PR Checklist

- [*] I have run `yarn test` and verified that the tests pass.
- [*] I have run `yarn generate` to generate and format code and docs.
- [*] The attribute is in a namespace (e.g. `nextjs.function_id`, not `function_id`)
- [*] I have used the correct value for `apply_scrubbing` (i.e. `manual` or `auto`. Use `never` only for values that should never be scrubbed such as IDs)


Lms24 (commented):
Generally no objections just a thought: Have we considered reusing `sentry.origin`? 

The only thing that speaks a bit against `sentry.metric.source` is that there's `sentry.span.source` and it has a [different meaning](https://getsentry.github.io/sentry-conventions/attributes/sentry/#sentry-span-source). Though we'll likely deprecate `sentry.span.source` anyway, so it's not too big a deal.

klochek:
> Generally no objections just a thought: Have we considered reusing `sentry.origin`?
> 
> The only thing that speaks a bit against `sentry.metric.source` is that there's `sentry.span.source` and it has a [different meaning](https://getsentry.github.io/sentry-conventions/attributes/sentry/#sentry-span-source). Though we'll likely deprecate `sentry.span.source` anyway, so it's not too big a deal.

For our (Relay's) purposes, we're passing through `sentry.origin` since it contains the originating web vital name, like "auto.http.browser.inp".  Perhaps we _could_ repurpose it on our end for the metric provenance, but I feel like it'd be safer not to.

logaretm:
While I don't exactly remember why we didn't consider re-using the `origin` initially, I think preserving it from the span origin could be desirable here, so a new attribute is needed to avoid having web vital spans with diverging origins. Does that make sense?

Lms24 (approved):
Makes sense, thx for explaining!

PR #481, opened 2026-07-10
feat(attributes): Add faas.invoked_* attributes

andreiborza:
Adds the `faas.*` attributes emitted by the vendored JS AWS SDK instrumentation.

- invoked_name
- invoked_provider
- invoked_region

OTel sem conv for `faas.*` can be found at: https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -5169,6 +5169,69 @@ export const FAAS_INVOCATION_ID = 'faas.invocation_id';
  */
 export type FAAS_INVOCATION_ID_TYPE = string;
 
+// Path: model/attributes/faas/faas__invoked_name.json
+
+/**
+ * The name of the invoked function. `faas.invoked_name`
+ *
+ * Attribute Value Type: `string` {@link FAAS_INVOKED_NAME_TYPE}
+ *
+ * Apply Scrubbing: manual
+ *
+ * Attribute defined in OTEL: Yes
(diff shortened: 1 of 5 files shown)


Lms24 (commented):
given that these attributes (according to otel) should be [equal](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/#faas-invoked-name:~:text=%5B2%5D%20faas.invoked_name%3A%20SHOULD%20be,region%20resource%20attribute%20of%20the%20invoked%20function.) to more general attributes: Can we set the `alias` field and link to them? 

No need to deprecate then, since they're also not deprecated in OTel. Though I wonder why OTel permits this duality 😅

Lms24 (approved):
As discussed offline: No need for the alias. These _should_ be the same value but it's not a bidirectional relationship.

andreiborza:
> given that these attributes (according to otel) should be [equal](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/#faas-invoked-name:~:text=%5B2%5D%20faas.invoked_name%3A%20SHOULD%20be,region%20resource%20attribute%20of%20the%20invoked%20function.) to more general attributes: Can we set the `alias` field and link to them?
> 
> No need to deprecate then, since they're also not deprecated in OTel. Though I wonder why OTel permits this duality 😅

I read that more as a "these should have equal value when set" but not necessarily that these are aliases of each other. As discussed, probably better to keep as is.

PR #491, opened 2026-07-13
feat(attributes): Add `sentry.relay.ingress` and `sentry.relay.pipeline`

loewenheim:
## Description

Names are open to discussion. The idea is that Relay sets these attributes on items it receives and processes in order to give us some insight into the path items take through Relay.

ref: INGEST-943

## PR Checklist

If an attribute was added:


cleptric (approved):

model/attributes/sentry/sentry__relay__pipeline.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "sentry.relay.pipeline",
+  "brief": "The pipeline an item went through in Relay.",
```
Dav1dde: Maybe:

```suggestion
  "brief": "An internal descriptor of which processing pipeline an item went through in Relay.",
```

Dav1dde (approved):

Lms24 (approved):

PR #492, opened 2026-07-13
feat(attributes): add search alias metadata

### LINKED ISSUE #85: Add units to conventions
All numeric conventions should have units

cleptric:
This PR (edited by @Lms24):

- adds a `search_alias` property to the attributes schema
  - `name`: main alias in sentry search
  - `type`: any narrowing search types (units) that are not the default primary `type`
    - worth noting: primary `type: 'double'` should convert to Sentry product type `number`. We decided to handle this in Sentry code rather than conventions
  - `deprecated_aliases`: List of deprecated aliases used and supported in the product. At this time, it's only `span.system`.
 - adds search_types entries to attributes that are currently registered as a `ResolvedAttribute` in the Sentry backend  
 - adds JS, Rust and Python codegen for `search_alias`, put onto the `ATTRIBUTE_METADATA` object
 - adjusts docs to show search alias information and update the search index with aliases for better searching
 
image


model/attributes/sentry/sentry__op.json
```diff
@@ -8,6 +8,17 @@
   "is_in_otel": false,
   "example": "http.client",
   "visibility": "public",
+  "search_aliases": [
+    {
+      "public_alias": "span.op",
+      "internal_name": "sentry.op",
+      "search_type": "string",
```
cleptric: Can likely make string the default, so not required.

schemas/attribute.schema.json
```diff
@@ -99,6 +99,63 @@
...
+        "additionalProperties": false,
+        "properties": {
+          "public_alias": {
+            "description": "The public name exposed in Sentry search",
+            "type": "string"
+          },
+          "internal_name": {
+            "description": "The internal attribute name used by the EAP search resolver",
+            "type": "string"
+          },
```
mjq: Will this always duplicate the attribute's `key`? It feels like it ought to, in which case I'd prefer we make mismatches impossible by removing this property.

schemas/attribute.schema.json
```diff
@@ -99,6 +99,63 @@
...
+            ]
+          },
+          "secondary_alias": {
+            "description": "Whether the alias is accepted for queries but omitted from autocomplete suggestions",
+            "type": "boolean"
+          },
+          "private": {
+            "description": "Whether the internal attribute should be hidden from public search results",
+            "type": "boolean"
+          },
```
mjq: Is this already covered by `visibility`? cc @nsdeschenes

schemas/attribute.schema.json
```diff
@@ -99,6 +99,63 @@
...
+          "dataset_mappings": {
+            "description": "Exact query column mappings for Sentry's span datasets",
+            "type": "object",
+            "additionalProperties": false,
+            "minProperties": 1,
+            "properties": {
+              "spans": {
+                "description": "The column expression in the indexed spans dataset",
+                "type": "string"
+              },
```
mjq: This dataset has been removed. Only EAP exists now.

schemas/attribute.schema.json
```diff
@@ -99,6 +99,63 @@
...
+              "eap": {
+                "description": "The column expression in the Events Analytics Platform dataset",
+                "type": "string"
+              }
```
mjq: This seems better left to the Sentry backend to compute. (What's sent to EAP isn't a string either, so I don't think this is helpful here anyway). We do need a way to record when the search types and internal types of an attribute mismatch, though. I'd rather expose an (optional) `internal_type` property for this (one of `string`, `boolean`, `double`, `integer`, or `array`) so that the backend can construct the correct RPC proto.

schemas/attribute.schema.json
```diff
@@ -99,6 +99,63 @@
...
+          },
+          "secondary_alias": {
+            "description": "Whether the alias is accepted for queries but omitted from autocomplete suggestions",
+            "type": "boolean"
+          },
+          "private": {
+            "description": "Whether the internal attribute should be hidden from public search results",
+            "type": "boolean"
+          },
+          "dataset_mappings": {
```
mjq: Given the two other two comments re: `spans` and `eap` I think the whole `dataset_mappings` can be removed.

buenaflor:
this closes https://github.com/getsentry/sentry-conventions/issues/85 right?

model/attributes/cache/cache__hit.json
```diff
@@ -8,6 +8,10 @@
   "is_in_otel": false,
   "example": true,
   "visibility": "public",
+  "search_alias": {
```
Lms24, line 11: q: Does it make sense to add a `search_alias` entry if neither type nor name differ from the primary definition? like in this case but probably in a lot of others, too

model/attributes/db/db__system.json
```diff
@@ -13,6 +13,10 @@
     "replacement": "db.system.name"
   },
   "visibility": "public",
+  "search_alias": {
+    "name": "db.system",
+    "aliases": ["span.system"]
```
Lms24, line 18: side-note/no action required: I guess we can't break this, since we have such a `ResolvedAttribute` but IMHO the aliases mapping makes very little sense 🤔

model/attributes/sentry/sentry__thread__id.json
```diff
@@ -1,13 +1,16 @@
 {
   "key": "sentry.thread.id",
-  "brief": "Current “managed” thread ID.",
+  "brief": "Current \u201cmanaged\u201d thread ID.",
```
Lms24, line 3: super-l:

```suggestion
  "brief": "Current \"managed\" thread ID.",
```

schemas/attribute.schema.json
```diff
@@ -153,6 +153,29 @@
...
+        "type": {
+          "description": "The type exposed by Sentry search. Defaults to string if omitted",
+          "enum": ["string", "boolean", "integer", "number", "byte", "currency", "millisecond", "percentage", "second"]
+        },
+        "aliases": {
+          "description": "Additional aliases accepted in search queries",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
```
Lms24, line 174: m: what's the difference between `search_alias.name` and `search_alias.aliases`? Is the intention purely to have more than one alias or are there more differences? 

If there are no differences, should we simplify to `names` which is always an Array? (I likely lack some context in the product but I checked `ResolvedAttribute` and I don't see more than one alias per instance. I do see the `secondary_alias` boolean though which might play a role here 🤔 )

If there is a difference, let's point this out in the schema description. Also let's consider changing the name because we already have a top-level `alias` field and this might get confusing, especially, because the top-level `alias` get's generated to `aliases` in JS and Python code. So if these are "secondary" aliases, let's maybe call it `secondary_aliases`?

Lms24:
I will rebase https://github.com/getsentry/sentry-conventions/pull/508 on this PR and probably repurpose it slightly: Instead of exposing purely a deprecation chain, the chains will include:

- the primary attribute name
- any search aliases (alias.name + alias.aliases or whatever definition we settle on (see https://github.com/getsentry/sentry-conventions/pull/492#discussion_r3735732801))
- the aliasing attributes
- the primary name and aliases of any replacing or deprecated attributes as long as their status is `backfill` or `normalize`.

schemas/attribute.schema.json
```diff
@@ -153,6 +153,29 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "name": {
+          "description": "The public name exposed in Sentry search",
+          "type": "string"
+        },
+        "type": {
+          "description": "The type exposed by Sentry search. Defaults to string if omitted",
+          "enum": ["string", "boolean", "integer", "number", "byte", "currency", "millisecond", "percentage", "second"]
```
Lms24: Do we need to include types that are identical to the primary type? My (obviously very limited) understanding is we only need to annotate attributes with "special" types that are narrower than the attribute type. Like`"byte", "currency", "millisecond", "percentage", "second"`.

model/attributes/cache/cache__hit.json
Lms24: discussed async: Yes, we can omit these

schemas/attribute.schema.json
Lms24: discussed async: secondary aliases are deprecated aliases, hence we settled on `deprecated_aliases`

schemas/attribute.schema.json
Lms24: discussed offline: We only need special, narrower types here. primitive types can be omitted.

nsdeschenes (approved):

PR #504, opened 2026-07-14
ci(release): show version in workflow run title

constantinius:
## Description

Sets the release workflow `run-name` to include the manually dispatched `version` input. This makes previous Release workflow runs show the released version directly in the GitHub Actions run list. See https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#run-name for reference

## Validation

- Reviewed `.github/workflows/release.yml` diff


Lms24 (approved):

Lms24:
Neat QoL improvement! Just to be clear, we still should check the actual changes to determine patch vs. minor

PR #505, opened 2026-07-15
feat(conventions): Support multiple examples

### LINKED ISSUE #209: Provide multiple examples for attributes
Right now, we can only provide one example for each attribute. Sometimes, attribute semantics can be more clearly explained with multiple examples. Therefore, it'd make sense to me to support an array of examples.

Lms24:
Add support for multiple attribute examples across validation, authoring, generated JavaScript and Python metadata, and documentation while preserving legacy `example` compatibility. This is helpful when one example alone isn't enough to illustrate the possible values.

Also added a few examples to `db.query.summary` which is such a case IMHO.

cc @mydea since we talked about this

Fixes #209

### DIFF
--- a/python/src/sentry_conventions/attributes.py
+++ b/python/src/sentry_conventions/attributes.py
@@ -5604,7 +5604,8 @@ class ATTRIBUTE_NAMES(metaclass=_AttributeNamesMeta):
     Defined in OTEL: Yes
     Visibility: public
     Aliases: url.template
-    Example: "/users/:id"
+    Example: "/users/:userID?"
+    Example: "my-controller/my-action/{id?}"
     """
 
     # Path: model/attributes/http/http__scheme.json
@@ -9143,7 +9144,9 @@ class ATTRIBUTE_NAMES(metaclass=_AttributeNamesMeta):
     Defined in OTEL: Yes
     Visibility: public
     Aliases: http.route
+    Example: "/users/{id}"
     Example: "/users/:id"
+    Example: "/users?id={id}"
     """
 
     # Path: model/attributes/url.json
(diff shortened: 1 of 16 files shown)


python/src/sentry_conventions/attributes.py
```diff
@@ -15930,8 +15942,12 @@ class ATTRIBUTE_NAMES(metaclass=_AttributeNamesMeta):
...
         example="/users/:id",
+        examples=["/users/:id", "my-controller/my-action/{id}", "/posts"],
         aliases=["url.template"],
         changelog=[
+            ChangelogEntry(
+                version="next", prs=[505], description="Added multiple examples"
+            ),
             ChangelogEntry(version="0.1.0", prs=[127]),
             ChangelogEntry(version="0.0.0"),
         ],
```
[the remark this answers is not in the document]
Lms24: I think you're drunk seer. nothing is swapper or out of sync.

scripts/types.ts
nsdeschenes, line 1: Wonder if it would be worth it to copy over the some of schemas from the content config and use them here then you can infer the types for example `AttributeValue` 👀

scripts/types.ts
```diff
@@ -21,6 +24,26 @@ export interface AttributeJson {
...
+  }
+  return undefined;
+}
+
+export function parseAttributeExamples(value: string): AttributeValue[] {
+  const examples = JSON.parse(value) as unknown;
+  if (!Array.isArray(examples) || examples.length === 0) {
+    throw new Error('Examples must be provided as a non-empty JSON array');
+  }
+  return examples as AttributeValue[];
```
nsdeschenes: And then you wouldn't have to cast =P

nsdeschenes (approved):

Lms24:
> Wonder if it would be worth it to copy over the some of schemas from the content config and use them here then you can infer the types for example AttributeValue

Good point. Seems like we have three type/schema definitions at the moment:

- schemas with JSON schema
- docs/src/content.config.ts
- scripts/types.ts

Would be nice if we could unify these or at least keep them better in sync. Definitely opens up room for divergence at the moment. I'm gonna think a bit about this a bit and tackle as a follow-up. Thanks for calling this out!

PR #508, opened 2026-07-15
feat(packages): Expose attribute key chains on `ATTRIBUTE_METADATA`

### LINKED ISSUE #496: Expose deprecation chains from conventions packages
Today, our SDKs sometimes have code paths in which we (have to) read span attributes to determine some information. A classic use case is OTel spans coming from frameworks (NextJS, SvelteKit) that need some sentry-SDK processing. A pain point here is that usually, this leads to SDK code like

```ts
const method = attributes['http.request.method'] || attributes['http.method'] || attributes['method'] 
```

where we need to check multiple attributes (the currently favoured one as well as previously deprecated variants) to reliably get the value we need. The most common use case is one instrumentation emitting deprecated attributes while another already emits the replacing one.

What we could do to future-proof ourselves is to have conventions expose a list of alias attributes that alias the queried attribute.

Sticking with the examples above, it would look somewhat like from conventions:

```ts
export const HTTP_REQUEST_METHOD = 'http.request.method';
export const HTTP_METHOD = 'http.method';
export const METHOD = 'method';

export const HTTP_REQUEST_METHOD_ALIASES = [HTTP_METHOD, METHOD];
export const HTTP_METHOD_ALIASES = [HTTP_REQUEST_METHOD, METHOD];
export const METHOD_ALIASES = [HTTP_REQUEST_METHOD, HTTP_METHOD]
```

Then, either in conventions or the package consumer (e.g. SDK), a helper could do something like

```ts
function getAttributeWithAliases(attributes: Record<string, unknown>, aliases: string[]) {
  return aliases.find(a => attributes[a]);
}

const method = getAttributeWithAliases(spanAttributes, HTTP_REQUEST_METHOD_ALIASES)
```

Which allows for a safe way of walking deprecation and alias chains, while keeping things reasonably tree-shakeable to avoid massive bundle size overhead.

Why not use `ATTRIBUTE_METADATA` which is already exposed?

This is a massive object containing all attributes in conventions. It would infeasibly increase bundle size of SDKs. More fine grained aliasing lists, like the one proposed, have much less of an impact (in the ballpark of what we do today).

Lms24:
Expose attribute key chains on the `ATTRIBUTE_METADATA` object under `ATTRIBUTE_METADATA[attributeKey].keys`.
 
A chain leads with the stable, non-deprecated key, so reading the chain from any key in the family always prefers the current key regardless of which key the caller uses. Within a chain the order is:

- the stable attribute, then its `search_alias.name` and `search_alias.deprecated_aliases`
- its non-deprecated aliases, each expanded the same way
- the deprecated attributes it replaces (alphabetically), each expanded the same way

Only `backfill` and `normalize` deprecations are taken into accound. A `transform` or `null` deprecation is not part of the chain since it signals a shift in semantics or attribute values.

### Potential Usage

`@sentry/conventions` consumers can make a bundle-size-safe lookup with this like so:

```ts
function getAttributeValueForKeys<T = unknown>(attributes: Record<string, unknown>, keys: string[]): T | undefined {
  // first defined attribute wins
  const foundKey = keys.find(k => attributes[k] != null)
  return foundKey ? attributes[foundKey] : undefined
}

const httpMethod = getAttributeValueForKeys(spanJson.attributes, ATTRIBUTE_METADATA['http.method'].keys);
```

As long as accesses to the metadata object are string literals (as opposed to variables), bundlers can tree-shake out the unused entries. #569 adds a more size-efficient lookup to query across all attributes without having to specify a set of keys.

closes #496


Lms24:
bugbot review

nsdeschenes (approved):
May be worth checking out that Seer review comment just to make sure everything's 🤌

Lms24:
> May be worth checking out that Seer review comment just to make sure everything's

Yup, good point. I still had a bug in the code that would add attributes unidirectionally to `keys` despite the deprecation status being `null` or `transform`. Should be fixed now.

PR #509, opened 2026-07-15
feat(attributes): Add deprecated messaging.destination_kind

andreiborza:
## What

Adds the legacy `messaging.destination_kind` attribute to the model, marked deprecated with no replacement.

- Model: `model/attributes/messaging/messaging__destination_kind.json`
- Regenerated JS (`MESSAGING_DESTINATION_KIND`) and Python exports via `yarn generate`.

## Why

Older OTel instrumentations still emit `messaging.destination_kind` (e.g. the aws-sdk SNS instrumentation the Sentry JS SDK ports for parity). OTel has since removed it in favor of `messaging.operation.type` + `messaging.destination.name`, so it's modeled here as deprecated (`_status: null`, no replacement).

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -10756,6 +10756,28 @@ export const MESSAGING_DESTINATION_CONNECTION = 'messaging.destination.connectio
  */
 export type MESSAGING_DESTINATION_CONNECTION_TYPE = string;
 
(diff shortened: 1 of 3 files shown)


Lms24 (approved):
>  now models the destination kind via messaging.operation.type and messaging.destination.na

These aren't similar enough to add them as replacements, correct? i think that's fine fwiw, since we couldn't backfill them anyway. Just checking.

andreiborza:
Yeah I think it's not similar enough.

PR #513, opened 2026-07-20
feat(attributes): Add browser.bfcache.* attributes

logaretm:
Registers the four `browser.bfcache.*` attributes emitted by the new `bfcacheMetricsIntegration` in the JS SDK (getsentry/sentry-javascript#22397).

They describe browser back/forward-cache health metrics.

| Attribute | Description |
| --- | --- |
| `browser.bfcache.outcome` | `hit`/`miss` of a back/forward navigation. |
| `browser.bfcache.reason` | A browser-reported not-restored reason (one per reason; a single miss can have several; Chromium-only via `notRestoredReasons`). |
| `browser.bfcache.frame` | Which frame the reason came from (`top`/`child`). |
| `browser.bfcache.not_restored_reason_count` | Number of reported reasons (`0` when the browser reported none, e.g. non-Chromium). |

All are Sentry-specific, so `is_in_otel: false` and PII is not applicable here.

### DIFF
--- a/model/attributes/browser/browser__bfcache__frame.json
+++ b/model/attributes/browser/browser__bfcache__frame.json
@@ -1,0 +1,18 @@
+{
+  "key": "browser.bfcache.frame",
+  "brief": "Which frame a back/forward cache not-restored reason originated from. 'masked' indicates a cross-origin subframe whose specific reason is hidden by the browser.",
+  "type": "string",
+  "apply_scrubbing": {
+    "key": "never"
+  },
+  "is_in_otel": false,
+  "visibility": "public",
+  "examples": ["top", "child", "masked", "unknown"],
+  "changelog": [
+    {
+      "version": "next",
+      "prs": [513],
+      "description": "Added browser.bfcache.frame attribute"
+    }
+  ]
+}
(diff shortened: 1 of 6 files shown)


model/attributes/browser/browser__bfcache__frame.json
```diff
@@ -0,0 +1,18 @@
+{
+  "key": "browser.bfcache.frame",
+  "brief": "Which frame in the page's frame tree a back/forward cache not-restored reason originated from: the top document or a child frame.",
+  "type": "string",
+  "apply_scrubbing": {
+    "key": "never"
```
Lms24:
```suggestion
    "key": "manual"
```

I 100% agree that Pii isnt a concern here but we should only set `never` if product functionality would break if the attribute value got scrubbed. Which I think isn't the case. It would break created BF-cache dashboards but users would have to do their part for this to happen. So I'd suggest we set `"manual"`. Meaning if users configure this field to be scrubbed (for whatever reason) it will be scrubbed. Otherwise, it will be left as-is. More details: https://develop.sentry.dev/engineering-practices/sentry-conventions/#before-opening-a-pr

Lms24 (commented):
The attribute names look very reasonable to me, so no objections to adding them from my end! We should adjust the `apply_scrubbing` behaviour though (see comment).

model/attributes/browser/browser__bfcache__frame.json
logaretm: Ah okay, thanks for the clarification!

logaretm: Updated in 8a15474

Lms24 (approved):

PR #522, opened 2026-07-23
feat(ops): Add `navigation.redirect` op and update name and description rules

Lms24:
## Description

Browser SDKs emit `navigation.redirect` for history/route changes triggered by a client router via a redirection chain rather than from a user initiated `navigation`. This PR adds the op and adjusts our name and description inference rules to use the same rules as for `navigation`.

## PR Checklist

If an attribute was added:


cleptric (approved):

Lms24:
Gonna merge this directly after the first approval. Two weeks of review time is more than enough for other code owners to weigh in.

PR #534, opened 2026-07-27
feat(attributes): Add `address`, deprecated in favor of `server.address`

alexander-alderman-webb:
## Description

Set in the transaction-based mode in Python's `socket` integration:

https://github.com/getsentry/sentry-python/blob/5204ffa36063c4dc85dd1be1f0b64c72e8fe0390/sentry_sdk/integrations/socket.py#L83

## PR Checklist

If an attribute was added:

### DIFF
--- a/model/attributes/address.json
+++ b/model/attributes/address.json
@@ -17,7 +17,7 @@
   "changelog": [
     {
       "version": "next",
-      "prs": [529],
+      "prs": [534],
       "description": "Added address attribute"
     }
   ]
--- a/model/attributes/address.json
+++ b/model/attributes/address.json
@@ -1,0 +1,24 @@
+{
+  "key": "address",
+  "brief": "The destination hostname or IP address for a TCP connection.",
+  "type": "string",
+  "apply_scrubbing": {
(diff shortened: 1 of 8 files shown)


model/attributes/address.json
```diff
@@ -0,0 +1,24 @@
...
+  "alias": ["server.address", "http.server_name", "net.host.name", "http.host", "server_name"],
+  "deprecation": {
+    "replacement": "server.address",
+    "reason": "Old namespace-less attribute, to be replaced with server.address for span-first future",
+    "_status": "backfill"
+  },
+  "changelog": [
+    {
+      "version": "next",
+      "prs": [529],
```
Lms24:
```suggestion
      "prs": [534],
```

Lms24 (approved):

model/attributes/address.json
alexander-alderman-webb: thanks https://github.com/getsentry/sentry-conventions/pull/534/changes/68a6b998373217f5d820969cf6efc543791ed443

PR #535, opened 2026-07-27
feat(attributes): Add `gcp_region`, deprecated in favor of `cloud.region`

alexander-alderman-webb:
## Description

Document an attribute set in the transaction-based Python SDK world:

https://github.com/getsentry/sentry-python/blob/5204ffa36063c4dc85dd1be1f0b64c72e8fe0390/sentry_sdk/integrations/gcp.py#L71

## PR Checklist

If an attribute was added:

### DIFF
--- a/model/attributes/gcp_region.json
+++ b/model/attributes/gcp_region.json
@@ -16,7 +16,7 @@
   "changelog": [
     {
       "version": "next",
-      "prs": [529],
+      "prs": [535],
       "description": "Added gcp_region attribute"
     }
   ]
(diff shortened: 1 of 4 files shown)


model/attributes/gcp_region.json
```diff
@@ -0,0 +1,23 @@
...
+  "examples": ["us-east-1"],
+  "alias": ["cloud.region"],
+  "deprecation": {
+    "replacement": "cloud.region",
+    "_status": "backfill"
+  },
+  "changelog": [
+    {
+      "version": "next",
+      "prs": [529],
```
Lms24:
```suggestion
      "prs": [535],
```

Lms24 (approved):

model/attributes/gcp_region.json
alexander-alderman-webb: https://github.com/getsentry/sentry-conventions/pull/535/changes/9999076180fde5f04acd269621bd240f9ca19e4b

PR #541, opened 2026-07-28
feat(attributes): Add cloudflare.agent attributes to identify agents

JPeer264:
## Description

This adds `cloudflare.agent.name` (name of the instance name) and `cloudflare.agent.class` (name of the class name - could be `Agent`, `AIChatAgent` or similar).

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -3564,6 +3564,50 @@ export const CLIENT_PORT = 'client.port';
  */
 export type CLIENT_PORT_TYPE = number;
 
+// Path: model/attributes/cloudflare/cloudflare__agent__class.json
+
+/**
+ * The class name of the Cloudflare agent that produced the span. `cloudflare.agent.class`
+ *
+ * Attribute Value Type: `string` {@link CLOUDFLARE_AGENT_CLASS_TYPE}
+ *
+ * Apply Scrubbing: manual
+ *
+ * Attribute defined in OTEL: No
+ * Visibility: public
+ *
+ * @example "MyBaseAgent"
+ * @example "AIChatAgent"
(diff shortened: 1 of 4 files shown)


Lms24 (commented):
Hmm I'm wondering if we can use more generic non-cloudflare-scoped attributes here?

- `cloudflare.agent.name` -> [`gen_ai.agent.name`](https://getsentry.github.io/sentry-conventions/attributes/gen_ai/#gen_ai-agent-name)
- `cloudflare.agent.class` -> `code.function.name` (assuming this is about the actual code class name). While a class is not a function, I think this could be close enough here. OTel is quite liberal in the [example](https://opentelemetry.io/docs/specs/semconv/registry/attributes/code/#code-function-name:~:text=Examples%3A,C%20function%3A%20fopen) values for this attribute.

alexander-alderman-webb:
Is `cloudflare.agent.class` needed to support any product feature?

JPeer264:
> Is cloudflare.agent.class needed to support any product feature?

No it is mainly to create more value on the span itself. Especially if there are more agents running at the same time.

> Hmm I'm wondering if we can use more generic non-cloudflare-scoped attributes here?

I like your ideas actually

> assuming this is about the actual code class name

It is about the class it extends from, so its parent class

Lms24:
Thanks! Ideally we don't add too many cloudflare(-or-any-other-lib/framework/plaform)-specific attributes when there are reasonable, more general replacements. Lmk if you wanna continue with any of these attributes :)

JPeer264:
> Ideally we don't add too many cloudflare(-or-any-other-lib/framework/plaform)-specific attributes

Totally agree. In this case I missed the alternatives. 

I'll close this as both of the alternatives are already in

PR #544, opened 2026-07-30
feat(op): Add `http.client.stream` op

msonnb:
## Description

Add `http.client.stream` for spans covering consumption of a streaming HTTP client response body

## PR Checklist

If an attribute was added:


Lms24 (approved):
thanks!

PR #545, opened 2026-07-31
feat(op): Add object storage span operations

msonnb:
## Description

Adds a new `object` op category with the following object storage related ops:
- `object.get`
- `object.head`
- `object.put`
- `object.delete`
- `object.list`
- `object.upload_part`
- `object.multipart_upload.abort`
- `object.multipart_upload.create`
- `object.multipart_upload.complete`

## PR Checklist

If an attribute was added:

### DIFF
--- a/model/op/object.json
+++ b/model/op/object.json
@@ -1,0 +1,42 @@
+{
+  "name": "object",
+  "description": "Object storage related spans represent operations on object storage systems such as Cloudflare R2, Amazon S3, and compatible services.",
+  "fields": [
+    {
+      "name": "object.get",
+      "description": "Retrieving an object from an object store."
+    },
+    {
+      "name": "object.head",
+      "description": "Retrieving the metadata of an object from an object store."
+    },
+    {
+      "name": "object.put",
+      "description": "Storing an object in an object store."
+    },
+    {
+      "name": "object.delete",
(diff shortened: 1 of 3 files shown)


model/op/object.json
```diff
@@ -0,0 +1,42 @@
...
+    {
+      "name": "object.delete",
+      "description": "Deleting an object from an object store."
+    },
+    {
+      "name": "object.list",
+      "description": "Listing objects in an object store."
+    },
+    {
+      "name": "object.upload_part",
```
msonnb, line 26: i think it would make sense to also put this under `object.multipart_upload.*`, but `object.upload_part` is how we currently set it in the JS R2 instrumentation (which we could change of course). Python doesn't seem to have any `object.*` ops at all.

Lms24 (approved):
All of these ops exist in [dev docs](https://develop.sentry.dev/sdk/telemetry/traces/span-operations/#web-server), so this is fine to add here. Thanks!

model/op/object.json
Lms24: yeah generally agreed but I think it's fine. this op is also mentioned in dev docs: https://develop.sentry.dev/sdk/telemetry/traces/span-operations/#web-server, so it's somewhat "official".

PR #551, opened 2026-08-04
feat(op): Add `router` span operation

msonnb:
## Description
Register the `router` operation and a new `routing` category for work performed by an application router, along with its span name inference rules.

## PR Checklist

If an attribute was added:


Lms24 (commented):
Let's add a `description` rule as well. As discussed, we can also include `url.path` as another fallback attribute for description (but not for name)

Lms24 (approved):
LGTM from my end! cc @mjq @nsdeschenes are you fine with the description and name rules?

PR #554, opened 2026-08-04
feat(op): Add `handler` op

msonnb:
## Description

Add `handler` op to the `web_server` category for spans that instrument http request handlers in server frameworks like express, fastify, etc.

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/op.ts
+++ b/javascript/sentry-conventions/src/op.ts
@@ -487,7 +487,7 @@ export const WEB_SERVER_MIDDLEWARE_SPAN_OP = 'middleware';
 /**
  * Handling of an incoming request by a web server route handler
  */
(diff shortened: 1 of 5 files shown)


Lms24 (commented):
Let's do `handler` instead please after discussing this with @cleptric. Thanks!

Lms24 (commented):
Thanks! I think we'll also need description and name rules for this op. I think we can use the following templates for these spans from checking the use cases in the SDK: 

Description:
1. `{http.route}`
2. `{url.path}`

Name:
1. `{http.route}`
2. `"Request handler`

Does that sound reasonable to you?

msonnb:
sound reasonable yes! added in https://github.com/getsentry/sentry-conventions/pull/554/commits/f23927c95b67dc414a11fc4c7117caf69102c10f

Lms24 (approved):
thanks!

PR #558, opened 2026-08-07
feat(ops): Add MCP ops

msonnb:
## Description

Adds ops and description/name rules for `mcp.server`, `mcp.notification.client_to_server`, and `mcp.notification.server_to_client`, as already used by the JS SDK.

## PR Checklist

If an attribute was added:


Lms24 (approved):

alexander-alderman-webb (approved):
We only have `mcp.server` in Python.
The other ops look strange but that's okay I guess.

PR #562, opened 2026-08-07
perf(js): Make attribute descriptions tree-shakeable

scttcper:
Sentry only uses the browser name and version descriptions in its frontend, but reading them from `ATTRIBUTE_METADATA` pulls in the entire table.

This moves documentation fields to an opt-in entry and adds named brief exports:

```ts
import {
  BROWSER_NAME_BRIEF,
  BROWSER_VERSION_BRIEF,
} from "@sentry/conventions/attributes/documentation";
```

With Rspack, that Sentry import drops from 240,110 bytes minified to 93 bytes. `ATTRIBUTE_DOCUMENTATION` remains available for build-time tooling that needs every description.


javascript/sentry-conventions/src/attributes.ts
```diff
@@ -19207,4194 +19188,2916 @@ export type AttributeName =
...
     visibility: 'public',
-    example: ['hello', 'world'],
     deprecation: {
       replacement: 'gen_ai.output.messages',
       status: 'backfill',
     },
-    changelog: [{ version: '0.1.0', prs: [65, 127] }, { version: '0.0.0' }],
   },
   'ai.response_format': {
-    brief: 'For an AI model call, the format of the response',
```
scttcper, line 19542: i guess we probably want the brief

scttcper:
i'm not sure this is the right way to go

Lms24:
@scttcper have you tried accessing `ATTRIBUTE_METADATA['specific.attribute_name']`? I believe bundlers should be smart enough to tree-shake the unused parts of the object if every access to the metadata object is done with a specific string (instead of `ATTRIBUTE_METADATA[someVariable]`.

That being said, if you prefer the individually exported constants, I'm happy to merge this. Bundle size of these is still a couple of bytes less than one entire metadata object entry.

PR #566, opened 2026-08-10
feat(attributes): Add `sentry.event.serialized_meta`

mjq:
Add an internal `sentry.event.serialized_meta` attribute for holding Relay's metadata (PII scrubbing, etc) for the contexts, extra, and breadcrumbs in the other `sentry.event.serialized_*` attributes. This is currently used in the span details pane for transaction-event-based transaction/segment spans, and therefore required for backwards compatibility.


mjq:
Related Relay PR: https://github.com/getsentry/relay/pull/6296

nsdeschenes (approved):

PR #571, opened 2026-08-13
fix(attributes): Add status `backfill` to deprecated attributes with renamed replacements

### LINKED ISSUE #570: Semantically equivalent deprecated attributes have deprecation status `null`
A bunch of attributes are:
1. deprecated
2. list a replacement
3. semantically equivalent to the replacement
4. have status `null`

I think this is an issue and makes cross-attribute-key lookups like we need to do in the sentry frontend (via #569 and #508) harder than necessary. My guess is, these attributes were added prior to backfilling/normalization but I might also lack context here.

Here's a list:

| Old | Replacement | Note |
| --- | --- | --- |
| `code.filepath` | `code.file.path`  | |
| `code.lineno` | `code.line.number`  | |
| `db.name` | `db.namespace`  | |
| `http.client_ip` | `client.address`  | |
| `http.flavor` | `network.protocol.version`  | |
| `http.scheme` | `url.scheme`  | |
| `http.server_name` | `server.address`  | |
| `http.status_code` | `http.response.status_code`  | |
| `http.url` | `url.full` | |
| `http.user_agent` | `user_agent.original` | |
| `net.host.ip` | `network.local.address` | |
| `net.host.name` | `server.address`  | |
| `net.host.port` | `server.port`  | |
| `net.peer.ip` | `network.peer.address`  | |
| `net.protocol.name` | `network.protocol.name`  | |
| `net.protocol.version` | `network.protocol.version`  | |
| `net.sock.host.port` | `network.local.port`  | |
| `net.sock.peer.addr` | `network.peer.address`  | Old was IP-only; still a valid value of the target |
| `net.sock.peer.port` | `network.peer.port`  | |
| `net.transport` | `network.transport`  | |
| `runtime.name` | `process.runtime.name`  | |
| `runtime.raw_description` | `process.runtime.description`  | Target brief states it is equivalent to `raw_description` |
| `runtime.version` | `process.runtime.version`  | |
| `sentry.browser.name` | `browser.name`  | |
| `sentry.browser.version` | `browser.version`  | |
| `sentry.user.email` | `user.email`  | |
| `sentry.user.geo.city` | `user.geo.city`  | |
| `sentry.user.geo.country_code` | `user.geo.country_code`  | |
| `sentry.user.geo.region` | `user.geo.region`  | |
| `sentry.user.geo.subdivision` | `user.geo.subdivision`  | |
| `sentry.user.id` | `user.id`  | |
| `sentry.user.ip` | `user.ip_address`  | |
| `sentry.user.username` | `user.name`  | |
| `url` | `url.full`  | |

I propose we set `backfill` or `normalize` to these attributes or call out why they should not be migrated

Lms24:
## Description

This PR applies the deprecation status `"backfill"` to a bunch of attributes that were straight renames but for unknown reasons had status `null`. This means, these attributes are now added to the deprecation chains on the `ATTRIBUTE_METADATA` object to allow for deprecation-safe attribute accesses.

closes https://github.com/getsentry/sentry-conventions/issues/570

## PR Checklist

If an attribute was added:

### DIFF
--- a/model/attributes/net/net__sock__host__addr.json
+++ b/model/attributes/net/net__sock__host__addr.json
@@ -9,7 +9,7 @@
   "example": "/var/my.sock",
   "alias": ["network.local.address", "net.host.ip"],
   "deprecation": {
-    "_status": null,
+    "_status": "backfill",
     "replacement": "network.local.address"
   },
   "visibility": "public",
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -11858,6 +11858,7 @@ export type NETWORK_CONNECTION_TYPE_TYPE = string;
  * Aliases: {@link NET_HOST_IP} `net.host.ip`, {@link NET_SOCK_HOST_ADDR} `net.sock.host.addr`
  *
  * @example "10.1.2.80"
+ * @example "/var/run/my.sock"
  */
 export const NETWORK_LOCAL_ADDRESS = 'network.local.address';
 
@@ -27469,6 +27470,7 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     isInOtel: true,
     visibility: 'public',
     example: '10.1.2.80',
+    examples: ['10.1.2.80', '/var/run/my.sock'],
     aliases: ['net.host.ip', 'net.sock.host.addr'],
     changelog: [{ version: '0.1.0', prs: [127] }, { version: '0.0.0' }],
   },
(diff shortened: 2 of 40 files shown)


model/attributes/net/net__sock__host__addr.json
```diff
@@ -9,7 +9,7 @@
   "example": "/var/my.sock",
   "alias": ["network.local.address", "net.host.ip"],
   "deprecation": {
-    "_status": null,
+    "_status": "backfill",
```
Lms24, line 12: given the example, the deprecation just being a rename doesn't seem to be 100% straight forward but OTEL also [lists](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#network-local-address) that `network.local.address` may hold `/var/my.sock`. I added it as an example in our definition of `network.local.address`

nsdeschenes (approved):

PR #572, opened 2026-08-24
feat(attributes): Add graphql.processing.type

andreiborza:
## Description

Adds `graphql.processing.type`, which says which part of GraphQL request processing a span represents: `parse`, `validate`, `execute`, `resolve` and so on. Adds it to the `graphql` span name templates as well, so the phases get a name of their own:

```
GraphQL {{graphql.operation.type}}
GraphQL {{graphql.processing.type}}
GraphQL Operation
```

GraphQL span names are low cardinality and carry the operation type at most, so a parse span and a validate span are otherwise indistinguishable. A parse span carries no GraphQL attribute at all today.

Both attributes are set on an operation span, and only `graphql.processing.type` on the phases, so the template order keeps `GraphQL query` for operations and gives `GraphQL parse` / `GraphQL validate` / `GraphQL resolve` to the rest.

Also replaces the `mutation` and `query` examples, which no template can produce since Sentry prefixes GraphQL span names with `GraphQL `, as the `otel_notes` on the same entry says. They predate that prefix.

The attribute name and its value list follow the GraphQL OpenTelemetry Working Group proposal in https://github.com/open-telemetry/semantic-conventions/pull/3515. That PR is still a working draft, so `is_in_otel` is `false` for now and should be flipped once it lands upstream. Note that the working group deliberately did not widen `graphql.operation.type` for this, since that attribute holds the GraphQL operation type (`query`, `mutation`, `subscription`) and applies only to spans that run an operation.

`@sentry/javascript` wants this for https://github.com/getsentry/sentry-javascript/pull/23542, where streamed GraphQL spans lose the operation name and the resolver field path from their names.

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -8595,6 +8595,30 @@ export const GRAPHQL_OPERATION_TYPE = 'graphql.operation.type';
  */
 export type GRAPHQL_OPERATION_TYPE_TYPE = string;
 
+// Path: model/attributes/graphql/graphql__processing__type.json
+
+/**
+ * The type of processing represented by this span. `graphql.processing.type`
+ *
+ * Attribute Value Type: `string` {@link GRAPHQL_PROCESSING_TYPE_TYPE}
+ *
+ * Apply Scrubbing: manual
+ *
+ * Attribute defined in OTEL: No
+ * Visibility: public
+ *
+ * @example "parse"
+ * @example "validate"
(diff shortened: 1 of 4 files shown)


mjq (approved):

Lms24 (approved):

Lms24 (approved):
attribute and name changes look good to me. IMHO these are all low-cardinality and the processing types were previously missing and would have caused us to fall back to the static string. So having this new rule should be fine.

PR #574, opened 2026-08-25
feat(attributes): Add HTTP body size and status text attributes

msonnb:
## Description

- Registers 4 deprecated HTTP attributes the JS SDK currently emits:
  - `http.request_content_length`
  - `http.request_content_length_uncompressed`
  - `http.response_content_length_uncompressed`
  - `http.status_text`,

- deprecates the existing `http.decoded_response_content_length`

- and registers the following replacements:
  - `http.request.body.size`
  - `http.request.body.decoded_size`
  - `http.response.body.decoded_size`
  - `http.response.status_text`

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -8959,6 +8959,9 @@ export type HTTP_CLIENT_IP_TYPE = string;
  * Attribute defined in OTEL: No
  * Visibility: public
  *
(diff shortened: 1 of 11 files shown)


Lms24 (approved):
LGTM! For other reviewers: The gist is that we want to continue sending compressed and uncompressed sizes and OTel only specifies compressed sizes. Therefore we introduce uncompressed size attributes that are in line with the new naming scheme.

PR #575, opened 2026-08-25
feat(span-names): Add `url.domain`rule for `http.client` spans

Lms24:
This PR proposes adding another name generation rule to `http.client` spans: If we cannot find a parameterized URL matching the first two rules (which is almost always the case in JS at least), we can construct the name based on 

```
{{http.request.method}} {{url.domain}}
```

to include at least a bit more information than just the method. Would appreciate product input if this is an acceptable level of cardinality!


mjq (approved):
For completeness: as mentioned in Slack there's a possibility of unlimited cardinality here via subdomains, but I think this is a more useful default, doesn't particularly have infrastructure implications, and users can override the names in an SDK hook if they have a use case with high-cardinality subdomains they'd prefer to be grouped. 👍

chargome (approved):

PR #577, opened 2026-08-25
chore(attributes): Add search alias for segment name

nsdeschenes:
Adding in `transaction` as the search alias for segment name.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -29608,7 +29608,7 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
   'sentry.segment.name': {
     brief: 'The segment name of a span',
     type: 'string',
-    keys: ['sentry.segment.name', 'sentry.transaction', 'transaction'],
+    keys: ['sentry.segment.name', 'transaction', 'sentry.transaction'],
     applyScrubbing: {
       key: 'manual',
     },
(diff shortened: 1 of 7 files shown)


mjq (changes requested):
I think this should be removed from `sentry.transaction` at the same time - having the same alias point to two different internal names seems likely to lead to confusion or bugs. (Adding a test to ensure that no search alias points to more than one attribute would be excellent)

nsdeschenes:
> I think this should be removed from `sentry.transaction` at the same time - having the same alias point to two different internal names seems likely to lead to confusion or bugs. (Adding a test to ensure that no search alias points to more than one attribute would be excellent)

Oops, thanks for the catch @mjq! I wonder if i should throw an error in the search meta data when there's more than one search alias found 🤔

mjq:
@nsdeschenes The way we've usually handled this is through attribute tests (`/test/attributes.test.ts`), e.g. [our alias tests](https://github.com/getsentry/sentry-conventions/blob/db559edf241ef156292af7b9e276a46632958b58/test/attributes.test.ts#L315). This kind of data consistency is a cross-cutting concern (affects all users, not just TS) and saves us having to have live assertions in all the libs.

nsdeschenes:
> @nsdeschenes The way we've usually handled this is through attribute tests (`/test/attributes.test.ts`), e.g. [our alias tests](https://github.com/getsentry/sentry-conventions/blob/db559edf241ef156292af7b9e276a46632958b58/test/attributes.test.ts#L315). This kind of data consistency is a cross-cutting concern (affects all users, not just TS) and saves us having to have live assertions in all the libs.

Ahhh okok, done! Should i remove the error being thrown during generation or is that one also alright to keep around?

mjq (approved):
@nsdeschenes Thanks! I personally don't think we still need `assertUniqueSearchAliases` + accompanying test in the generator if we have the search alias uniqueness test for attributes, but up to you :+1:

Lms24 (approved):
thanks for adjusting the tests! Looks good to me and I'm fine with also throwing during generation. No strong opinions as long as _something_ fails 😅

PR #581, opened 2026-08-27
feat(attributes): Add legacy AMQP and Redis span attributes (deprecated)

JPeer264:
Adds the seven legacy AMQP `messaging.*` attributes and `db.connection_string`, which the JavaScript SDK stopped emitting in v11. Without a definition here, nothing tells a user what replaced them.

`db.connection_string` has no single replacement: the connection is now described by `server.address` and `server.port`. It therefore carries a reason instead of a `replacement`, and gets no alias, so that no two-target alias group is created. (is there any protocol on how to treat these?)

`messaging.protocol`, `messaging.protocol_version` and `messaging.url` join existing alias groups, so the other members of those groups gain the new key as well.

Deprecated by https://github.com/getsentry/sentry-javascript/pull/23324

### DIFF
--- a/model/attributes/db/db__connection_string.json
+++ b/model/attributes/db/db__connection_string.json
@@ -1,0 +1,22 @@
+{
+  "key": "db.connection_string",
+  "brief": "The connection string used to connect to the database.",
+  "type": "string",
+  "apply_scrubbing": {
+    "key": "auto"
+  },
+  "is_in_otel": false,
+  "examples": ["redis://localhost:6379"],
+  "deprecation": {
+    "_status": null,
+    "reason": "This attribute is deprecated. The connection is described by server.address and server.port instead, so the value cannot be copied to a single replacement attribute."
+  },
+  "visibility": "public",
+  "changelog": [
+    {
+      "version": "next",
+      "prs": [581],
+      "description": "Added db.connection_string attribute"
+    }
+  ]
+}
--- a/model/attributes/db/db__connection_string.json
+++ b/model/attributes/db/db__connection_string.json
@@ -5,7 +5,7 @@
   "apply_scrubbing": {
     "key": "auto"
   },
-  "is_in_otel": false,
+  "is_in_otel": true,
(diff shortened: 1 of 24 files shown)


model/attributes/db/db__connection_string.json
```diff
@@ -0,0 +1,22 @@
+{
+  "key": "db.connection_string",
+  "brief": "The connection string used to connect to the database.",
+  "type": "string",
+  "apply_scrubbing": {
+    "key": "auto"
+  },
+  "is_in_otel": false,
```
Lms24: l: This is in OTel and also deprecated: https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-connection-string

```suggestion
  "is_in_otel": true,
```

model/attributes/db/db__connection_string.json
```diff
@@ -0,0 +1,22 @@
...
+  "examples": ["redis://localhost:6379"],
+  "deprecation": {
+    "_status": null,
+    "reason": "This attribute is deprecated. The connection is described by server.address and server.port instead, so the value cannot be copied to a single replacement attribute."
```
Lms24, line 12: good call to not add a replacement here, since it's two attributes that make up for this one. Thanks!

Lms24 (approved):
Thanks a lot for adding these!

model/attributes/db/db__connection_string.json
JPeer264: nice catch

model/attributes/db/db__connection_string.json
```diff
@@ -0,0 +1,22 @@
+{
+  "key": "db.connection_string",
+  "brief": "The connection string used to connect to the database.",
+  "type": "string",
+  "apply_scrubbing": {
+    "key": "auto"
+  },
+  "is_in_otel": true,
```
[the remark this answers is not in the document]
JPeer264: https://github.com/getsentry/sentry-conventions/pull/581#discussion_r3871929011

PR #582, opened 2026-08-27
feat(attributes): Add legacy gen_ai cache token attributes (deprecated)

JPeer264:
Adds `gen_ai.usage.cache_creation_input_tokens` and `gen_ai.usage.cache_read_input_tokens`, which the JavaScript SDK stopped emitting in v11 in favour of the dotted `gen_ai.usage.cache_*.input_tokens` names.

Both join the existing alias groups that already hold the Vercel AI spellings `gen_ai.usage.input_tokens.cache_write` and `gen_ai.usage.input_tokens.cached`.

Deprecated by https://github.com/getsentry/sentry-javascript/pull/22811


vgrozdanic (commented):
we need to first update the relay code for cost calculation before we can merge this, we currently never read `gen_ai.usage.cache_creation.input_tokens` in relay, and this change would make relay rewrite all the older fields to this new field `gen_ai.usage.cache_creation.input_tokens`.

This is a good change, we just need to first catch up with it in other places (and product potentially).

I see that OTEL has changed this last week:  https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/changelog.d/440.breaking.md?plain=1#L1

vgrozdanic (approved):

vgrozdanic:
> we need to first update the relay code for cost calculation before we can merge this, we currently never read `gen_ai.usage.cache_creation.input_tokens` in relay, and this change would make relay rewrite all the older fields to this new field `gen_ai.usage.cache_creation.input_tokens`.
> 
> This is a good change, we just need to first catch up with it in other places (and product potentially).
> 
> I see that OTEL has changed this last week: https://github.com/open-telemetry/semantic-conventions-genai/blob/67dff024110be5bd9f318006e733f4078e0f4c97/changelog.d/440.breaking.md?plain=1#L1

Relay will handle this quite well already, i wasn't aware that we don't have to worry about this, so this change is safe

Lms24 (approved):

PR #585, opened 2026-08-27
feat(attributes): Add `service.namespace`

JPeer264:
`service.namespace` is an OpenTelemetry resource attribute that SDKs emit alongside the already documented `service.name` and `service.version`, but it was never defined here.

It is added as a live attribute, not as a deprecated one. The JavaScript SDK stopped collecting OpenTelemetry resources in v11 and so no longer emits any `service.*` attribute, but that is a JavaScript-only change and other SDKs still send them. Deprecating the `service.*` family repo-wide would therefore be wrong.

Prompted by https://github.com/getsentry/sentry-javascript/pull/23013


JPeer264:
I don't think this is relevant.

PR #586, opened 2026-08-27
feat(attributes): Add `url.path.params.<key>` (deprecated) in favor of `url.path.parameter.<key>`

JPeer264:
The JavaScript SDK's TanStack Router integration emitted `url.path.params.<key>` and removed it in v11. It holds the same values as `url.path.parameter.<key>`, so it joins that attribute's alias group next to the existing `params.<key>`.

The original PR said `This PR removes the undocumented` so it would be interesting if this PR would still apply @Lms24 

Deprecated by https://github.com/getsentry/sentry-javascript/pull/23506

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -13426,7 +13426,7 @@ export type OTEL_STATUS_DESCRIPTION_TYPE = string;
  *
  * Has Dynamic Suffix: true
  *
- * Aliases: {@link URL_PATH_PARAMETER_KEY} `url.path.parameter.<key>`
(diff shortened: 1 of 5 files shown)


loewenheim (approved):

Lms24 (approved):

Lms24:
This change LGTM!

> The original PR said This PR removes the undocumented so it would be interesting if this PR would still apply @Lms24

sorry, which PR is that?

JPeer264:
> sorry, which PR is that?

The one which is linked below: https://github.com/getsentry/sentry-javascript/pull/23506
Just wondering if undocumented attributes should still be deprecated

PR #587, opened 2026-08-27
feat(attributes):  Name the replacements for the deprecated `http.target`

JPeer264:
`http.target` carried the path, the query string and the fragment in one value, so it cannot be renamed onto a single replacement. v11 of the JavaScript SDK splits it into `url.path`, `url.query` and `url.fragment`.

~Adds the `http_target_to_url_path_and_query` transformation and sets `http.target` to `_status: "transform"`.~

No aliases are added on purpose. An alias group is symmetric, so aliasing `http.target` to both `url.path` and `url.query` would declare those two equivalent to each other. This matches how `gen_ai.request.messages` is modelled.

Replaced by https://github.com/getsentry/sentry-javascript/pull/23575

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -10308,7 +10308,7 @@ export type HTTP_STATUS_TEXT_TYPE = string;
  * Attribute defined in OTEL: Yes
  * Visibility: public
  *
- * @deprecated Use {@link URL_PATH} (url.path) instead - This attribute is being deprecated in favor of url.path and url.query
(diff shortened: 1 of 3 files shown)


Lms24 (commented):
As discussed, we can just drop the transformation here and make this a hard replacement. With deprecation status `null`. Since multiple attributes replace this one, no need to add a `replacement` specifically but we just document it in the deprecation reason.

JPeer264:
@Lms24 the transformation is gone 🪄

Lms24 (approved):
thanks for adjusting!

PR #588, opened 2026-08-27
feat(attributes): Complete `net.*` aliases and transform `net.transport`

JPeer264:
Following were missing:

- `net.peer.name` replaces onto `server.address` but was not part of that alias group. It now joins it, like `http.host`, which is ambiguous between server and client in the same way.
- `net.sock.peer.port` replaces onto `network.peer.port` and neither named the other. The equivalent host pair, `net.sock.host.port` and `network.local.port`, already does.
- `net.transport` changes its values from `ip_tcp` and `ip_udp` to `tcp` and `udp`, so it needs a transformation rather than a plain rename. Adds `net_transport_to_network_transport`.

`net.host.port` and `net.peer.port` are left untouched. Both correctly replace onto `server.port`, while `network.local.port` and `network.peer.port` descend from the `net.sock.*` attributes, per the OpenTelemetry split.

Replaced by https://github.com/getsentry/sentry-javascript/pull/23301

### DIFF
--- a/model/attributes/net/net__transport.json
+++ b/model/attributes/net/net__transport.json
@@ -9,11 +9,16 @@
   "example": "tcp",
   "alias": ["network.transport", "mcp.transport"],
   "deprecation": {
-    "_status": "backfill",
-    "replacement": "network.transport"
+    "_status": null,
+    "replacement": "network.transport",
+    "reason": "This attribute is being deprecated in favor of network.transport. The values change from ip_tcp and ip_udp to tcp and udp, so the old value cannot be copied over."
   },
   "visibility": "public",
   "changelog": [
+    {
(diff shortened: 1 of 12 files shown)


model/attributes/net/net__transport.json
Lms24, line 1: To avoid the transformation here, my pragmatic take would be we just set status `'backfill'` and live with the slight deviations. cc @mjq / @nsdeschenes is that alright with you? 

Alternative would be `status: null`, which is probably also fine, given we don't seem to depend on the attributes in the frontend.

Lms24 (commented):
We need to address the transformation. The other changes look good to me, thanks!

Lms24 (approved):

PR #589, opened 2026-08-27
feat(attributes): Replace `fs_error` in favor of `error.type`

JPeer264:
`fs_error` held the full file system error message, for example `ENOENT: no such file or directory, open '/tmp/missing.txt'`. v11 of the JavaScript SDK reports the syscall error code on `error.type` instead, so the value has to be shortened, not just renamed.

Adds the `fs_error_to_error_type` transformation and sets `fs_error` to `_status: "transform"`.

No alias is added, matching how `gen_ai.request.messages` is modelled: an alias declares two names for the same value, which is no longer true once the value changes.

Replaced by https://github.com/getsentry/sentry-javascript/pull/23401

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -6394,6 +6394,8 @@ export type ENVIRONMENT_TYPE = string;
  * Attribute defined in OTEL: Yes
  * Visibility: public
  *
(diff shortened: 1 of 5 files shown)


JPeer264:
~we don't do transforms~ (Edit: but we keep the deprecation)

Lms24 (approved):
thanks! (PR title needs a small adjustment but otherwise lgtm!)

PR #590, opened 2026-08-27
feat(attributes): Complete the `gen_ai.tool.definitions` alias group

JPeer264:
`gen_ai.request.available_tools`, `ai.tools` and `ai.prompt.tools` all replace onto `gen_ai.tool.definitions`, but none of the four named the others. `gen_ai.request.available_tools` even carried an empty `alias` array and `ai.tools` had none at all, which is why the symmetry test stayed green. All four now list each other.

Replaced by https://github.com/getsentry/sentry-javascript/pull/22820


JPeer264:
how did that merge?

JPeer264:
stacks are confusing if you mess with them. It did not land on main.

PR #598, opened 2026-08-28
fix: Force use of Socket PyPI

sentrivana:
## Description

We should make it explicit that we want to use our Socket PyPI instance in `uv.lock`.

## PR Checklist

If an attribute was added:


Lms24 (approved):
Any chance this has something to do with socket.dev changes on our machines? Looks fine to me to merge this PR because I also noticed the lockfile getting changed locally but probably worth raising in discuss-security?

sentrivana:
Checking with Security folks before merging.

python/pyproject.toml
```diff
@@ -12,6 +12,10 @@ Homepage = "https://sentry.io"
 Repository = "https://github.com/getsentry/sentry-conventions.git"
 Issues = "https://github.com/getsentry/sentry-conventions/issues"
 
+[[tool.uv.index]]
+url = "https://pypi.org/simple"
```
oioki:
```suggestion
url = "https://sfw.security.sentry.io/pypi/simple"
```

Lms24 (approved):
thanks for following up on this!

PR #600, opened 2026-08-28
feat(attributes): Add browser.navigation.type, move navigation.* to router.navigation.*

logaretm:
Adds `browser.navigation.type` for web vital spans: `navigate`, `reload`, `prerender`, `bfcache`, `soft-navigation`. Mirrors what the web-vitals library reports as `navigationType`.

Also moves the `navigation.*` attributes to `router.navigation.*` so the router's own navigation type stops competing with the browser's. `navigation.type`, `navigation.origin` and `navigation.route.id` are deprecated with plain backfills, and the sveltekit aliases now point at the new names. These are straight renames so there's no transform involved.

The new attribute is going to be used to distinguish the various web vital signals so users can filter out specific (faster) types of navigations.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -3240,6 +3240,31 @@ export const BROWSER_NAME = 'browser.name';
  */
 export type BROWSER_NAME_TYPE = string;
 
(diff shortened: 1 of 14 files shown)


Lms24 (approved):
Looks good to me!

PR #603, opened 2026-08-31
chore(descriptions): Fix `http.client` span description example

Lms24:
## Description

`http.client` span descriptions call for the full URL as a description and not for just the path. The example incorrectly only included the path.

Side-note: Unfortunately, span descriptions will change slightly once they're inferred because SDKs had a dynamic logic to build `http.client` span descriptions:

- for same-origin requests, we'd include the relative URL (`/users/123?expand=full`)
- for 3rd-party origins, we'd include the full URL (`https://example.com/123?expand=full`)

## PR Checklist

If an attribute was added:

### DIFF
--- a/model/description/http.json
+++ b/model/description/http.json
@@ -6,7 +6,7 @@
       "brief": "Operations that represent outgoing HTTP requests.",
       "ops": ["http.client", "http.client.stream"],
       "templates": ["{{http.request.method}} {{url.full}}"],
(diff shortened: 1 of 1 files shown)


nsdeschenes (approved):

Lms24:
Gonna merge this for correctness but we likely need a larger conversation for the side-note in the PR

PR #609, opened 2026-09-02
fix(attributes): Change deprecation status from `backfill` to `normalize`

Lms24:
Changes the deprecation status of the attributes marked `backfill` in #571 to `normalize`.

Backfilling these straight renames means the value is stored under both the old and the new key. That increased log and metric sizes. Normalizing ensures a large value like `url.full` only ends up once in the stored metric.

Two attributes from #571 are untouched: `net.transport` (already reverted to `null` in #588) and `network.local.address` (that PR only changed its example).

@ Reviewers: If you prefer a more fine-grained change instead, I'm also happy to drop the bulk change.

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -23210,7 +23210,7 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
     example: '/app/myapplication/http/handler/server.py',
     deprecation: {
       replacement: 'code.file.path',
-      status: 'backfill',
+      status: 'normalize',
     },
     aliases: ['code.file.path'],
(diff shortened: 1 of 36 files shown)


Lms24:
It turns out, attributes being backfilled do not contribute to the billed log/metric size, as relay takes a size snapshot prior to backfilling. Therefore, I'm going to close the PR, since I slightly prefer the backfilling due to transparency. We can re-open whenver we need to!

loewenheim:
To expand a bit on how this works: the first _trusted_ Relay (i.e. in our infrastructure) that receives the logs (metrics, &c.) calculates the size before any processing and saves it in a header. This is the billing relevant size. Subsequent Relays don't update this header if it's present, so the size _as received_ is always what we use for billing no matter what transformations we do on the attributes.

PR #610, opened 2026-09-02
feat(search-attributes): Add search attribute names

nsdeschenes:
## Description

Generate typed `SEARCH_*` constants for Sentry search names so consumers can avoid hardcoded strings, and publish the search-specific API from its own module:

```ts
import {
  ATTRIBUTE_SEARCH_METADATA,
  SEARCH_TRANSACTION,
  type AttributeSearchName,
} from '@sentry/conventions/attributes/search';
```

## What changed

- Generate search constants, `AttributeSearchName`, search metadata types, and `ATTRIBUTE_SEARCH_METADATA` into a dedicated `search.ts` file instead of `attributes.ts`.
- Add the `@sentry/conventions/attributes/search` package export, build entry, and compatibility shims for older bundlers and TypeScript resolvers.
- Keep preferred search aliases as the current constants, such as `SEARCH_TRANSACTION = 'transaction'`.
- Continue emitting deprecated search names derived from attribute chains, marking them deprecated in favor of the preferred search alias where applicable.
- Keep current constants for attributes without an explicit search alias.
- Normalize away a leading `sentry.` in generated search names, so `sentry.transaction` and `transaction` produce only `SEARCH_TRANSACTION = 'transaction'`.
- Preserve actual storage keys in `canonicalName` and `deprecationChain` metadata without exposing duplicate `_SEARCH_*` constants.
- Preserve underscore distinctions so names such as `ai.model_id` and `ai.model.id` cannot collide.
- Keep `ATTRIBUTE_SEARCH_METADATA` keyed by `search_alias.name ?? attribute.key`.

## Test plan

- `yarn test:js` — 9,224 tests pass
- `yarn lint:types:js`
- `yarn lint:js`
- `yarn build:js`
- Verified a runtime import from `@sentry/conventions/attributes/search`


javascript/sentry-conventions/src/attributes.ts
```diff
@@ -33036,467 +32943,5617 @@ export const ATTRIBUTE_METADATA: Record<AttributeName, AttributeMetadata> = {
...
+
+/**
+ * Search name for {@link AI_MODEL_PROVIDER}. `ai.model.provider`
+ */
+export const SEARCH_AI_MODEL_PROVIDER = 'ai.model.provider';
+
+/**
+ * Search name for {@link _AI_MODEL_ID}. `ai.model_id`
+ */
+export const _SEARCH_AI_MODEL_ID = 'ai.model_id';
```
Lms24: are some consts like this one intentionally prefixed with `_`?

nsdeschenes: This was the 🤖 's initial attempt at resolving colliding names, this has been corrected.

scripts/generate_attributes.ts
```diff
@@ -1264,9 +1295,126 @@ function generateMetadataDict(
...
+  // Keys without a search alias still get a const; aliases and chain names already claimed win.
+  for (const { searchKey, preferredAttribute } of searchEntries) {
+    if (preferredAttribute.attributeJson.search_alias) {
+      continue;
+    }
+    if (searchNameConstantsByName.has(searchKey)) {
+      continue;
+    }
+    addSearchNameConstant(searchKey, false, preferredAttribute, searchKey);
+  }
```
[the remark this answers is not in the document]
nsdeschenes: This has been fixed up.

scripts/generate_attributes.ts
```diff
@@ -455,6 +475,35 @@ function getConstantName(key: string, isDeprecated: boolean): string {
...
+function getSearchConstantNameInner(searchKey: string): string {
+  // Original `_` becomes `__` first so `ai.model_id` → `SEARCH_AI_MODEL__ID` and
+  // does not collide with `ai.model.id` → `SEARCH_AI_MODEL_ID`.
+  return stripSentrySearchPrefix(searchKey)
+    .replaceAll('<', '')
+    .replaceAll('>', '')
+    .replaceAll('_', '__')
+    .replaceAll('.', '_')
+    .replaceAll('-', '_')
+    .toUpperCase();
```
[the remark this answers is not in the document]
nsdeschenes: Gonna accept this for the time being.

scripts/generate_attributes.ts
```diff
@@ -455,6 +475,35 @@ function getConstantName(key: string, isDeprecated: boolean): string {
...
+}
+
+function getSearchConstantNameInner(searchKey: string): string {
+  // Original `_` becomes `__` first so `ai.model_id` → `SEARCH_AI_MODEL__ID` and
+  // does not collide with `ai.model.id` → `SEARCH_AI_MODEL_ID`.
+  return stripSentrySearchPrefix(searchKey)
+    .replaceAll('<', '')
+    .replaceAll('>', '')
+    .replaceAll('_', '__')
+    .replaceAll('.', '_')
```
mjq: So this is pretty pedantic but could we consider flipping these?


```suggestion
    .replaceAll('_', '_')
    .replaceAll('.', '__')
```

`.` separates a level, and `_` separates a multi-word name inside that level. So `.` is a "heavier" split than `_`. I think the constants read better if names stay closer together - so e.g. keep "step functions" together as `SEARCH_AWS__STEP_FUNCTIONS__ACTIVITY__ARN` instead of `SEARCH_AWS_STEP__FUNCTIONS_ACTIVITY_ARN`.

FWIW this also would make it [align with Relay](https://github.com/getsentry/relay/blob/ee5277db77b6388724677bccb14f9433a6b04ead/relay-conventions/build/attributes.rs#L288-L292).

scripts/generate_attributes.ts
```diff
@@ -455,6 +475,35 @@ function getConstantName(key: string, isDeprecated: boolean): string {
   return constantName;
 }
 
+function stripSentrySearchPrefix(searchKey: string): string {
+  const prefix = 'sentry.';
+  return searchKey.startsWith(prefix) ? searchKey.slice(prefix.length) : searchKey;
```
mjq: TBH I think I'd rather handle these by putting search aliases on the attributes instead of a blanket policy here - having this enforced by the code generator feels like the wrong place compared to the model JSON 👀

scripts/generate_attributes.ts
nsdeschenes: Sounds good to me!

scripts/generate_attributes.ts
nsdeschenes: Okay, gonna remove this section for now, and roll without them and we can move to expanding the use of  visibility or providing proper search aliases when needed.

Lms24 (approved):

PR #620, opened 2026-09-03
feat(names): Add span name and description rules for `ui.*` spans

Lms24:
WIP


Lms24:
Superseded by #643, which adds the same `ui.*` name/description rules with updated templates (`ui.element.target`, route-id-first action names, split lifecycle fallbacks).

PR #625, opened 2026-09-04
fix(attributes): Set `db.query.text` scrubbing to `auto`

loewenheim:
## Description

[OTEL docs](https://opentelemetry.io/docs/specs/semconv/db/database-spans/#sanitization-of-dbquerytext) for the attribute say it should be scrubbed, and I tend to agree.

## PR Checklist


s1gr1d (approved):

Lms24 (approved):
So technically, SDKs should already sanitize this but I know of too many cases where we don't do this correctly. So +1 for now. Slight concern is span descriptions being scrubbed but we can revisit if it comes up

PR #626, opened 2026-09-07
chore(names): Clarify GraphQL OTel divergence note

andreiborza:
## What
The GraphQL span name OTel note now shows an example without an operation name.

## Why
Operation names are high-cardinality, so span names keep only the operation type.


andreiborza:
Otel does warn that these could be high-cardinality, but in practice we've never seen that. I also tried looking through public github repos to see if users define dynamic operations and didn't find any. Happy to close/revise if anyone has more experience with GraphQL and thinks this is a terrible idea.

Lms24 (commented):
I'm generally fine with this and happy to approve but I feel like this needs a product-sign-off. 

Just one technical suggestion: Let's add a sentence about diverging from the OTel naming schema in `otel_notes`.

andreiborza:
After discussions in slack, we came to the conclusion that these could be indeed high-cardinality, especially for public facing graphql api.

So I reworked this PR to just fix the wrong example in the conventions.

Lms24 (approved):

PR #632, opened 2026-09-09
chore(attributes): Clarify parameter name casing

Lms24:
Came up via https://github.com/getsentry/sentry-javascript/pull/24222

Makes it more explicit that we should follow established name patterns rather than force-normalize them, giving a couple of examples. The snake_case recommendation still stands, for example when we make up our own multi-word attributes like `*.web_vitals.*`.


nsdeschenes (approved):
Makes sense to me 😄

PR #636, opened 2026-09-10
feat(attributes): Add `sentry.app.url.domain` attribute

Lms24:
## Description

Today, all/most of our `url`, `server` or `http` attributes are interchangeably used to describe incoming and outgoing requests. This is (mostly) in line with OTel's definition. To differentiate, OTel uses `span.kind` and we use `sentry.op` (as well as `sentry.kind` for OTLP AFAIK). 

**Why do we need yet another URL attribute then?**

This PR adds a new attribute to serve a specific purpose: Provide the URL of the instrumented application on _every_ span, regardless of its kind/op. We need this to be able to make Sentry's "filter out events from localhost" inbound filter work. This includes `http.client` (outgoing request) spans. Their `url.*` attributes point to the request URL, so we can't set them to the application's URL. There are other spans like this, where we have similar issues (resource spans, db spans to a degree, etc). 

**Why the `sentry.*` Prefix?** 

Initially I wanted to go with `app.url.domain`, but then saw that OTel [defines](https://opentelemetry.io/docs/specs/semconv/registry/attributes/app/) the `app` namespace as _"attributes related to client-side applications"_. So I'd prefer not "misusing" this namespace for server-side applications. 

**Why this specific name?**

- `url.domain` already exists. Prefixing it with `sentry.app.` means we re-use an established pattern for a more specific purpose.
- We need some kind of prefix but I'm happy to change it if anyone has a better idea

**Alternatives considered?**

- Having two attributes, one for server, one for client. 
  - would work with inbound filters
  - For example: `browser.document.url.domain` (something [similar is in O](https://opentelemetry.io/docs/specs/semconv/registry/attributes/browser/#browser-document-url-full)Tel) and `http.server.url.domain` for the server-side. 
  - There's no precedence for this in OTel for the server side, and no server namespace exists for this, that isn't also used in e.g. browser-side outfoing request spans. 
  - That being said, this is still an option and we can also use this if anyone prefers
- A very specific "filtering attribute", something like `sentry.inbound_filter.url.domain`
  - Also an option I'm not opposed to
  - The counter argument is, attributes should describe a concept/semantic, not the purpose for what they are potentially used. 

## PR Checklist

If an attribute was added:

### DIFF
--- a/javascript/sentry-conventions/src/attributes.ts
+++ b/javascript/sentry-conventions/src/attributes.ts
@@ -14564,6 +14564,30 @@ export const SENTRY_ACTION = 'sentry.action';
  */
 export type SENTRY_ACTION_TYPE = string;
 
+// Path: model/attributes/sentry/sentry__app__url__domain.json
+
+/**
+ * The domain the instrumented app is running on. For browsers, that's the current window's url. For server-side applications, the domain the application is deployed on. Importantly, this attribute MUST NOT be used to describe urls of outgoing requests. `sentry.app.url.domain`
+ *
+ * Attribute Value Type: `string` {@link SENTRY_APP_URL_DOMAIN_TYPE}
+ *
+ * Apply Scrubbing: manual
+ *
+ * Attribute defined in OTEL: No
+ * Visibility: public
+ *
+ * @example "localhost"
+ * @example "api.project.com"
+ * @example "127.0.01"
+ * @example "::1"
+ */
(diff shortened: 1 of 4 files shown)


buenaflor (approved):
lgtm, I don't have a strong opinion on the naming but given the constraints I think it is the "lesser evil" like you described

Lms24:
closing in favour of #646

PR #641, opened 2026-09-21
feat(attributes): Add the INP target and interaction type attributes

logaretm:
The element and interaction type INP was reported on only exist in the span's name and op, I think it's a bit of a guess-work and is subject to breakage if we decide to rename span name/ops, an explicit attributes makes the most sense here.

I propose defining `browser.web_vital.inp.target` and `browser.web_vital.inp.interaction_type` as attributes the SDK can send optionally when they are known.


model/attributes/browser/browser__web_vital__inp__target.json
Lms24, line 1: Generally fine with this but to think about: We have a `browser.web_vital.lcp.element` attribute. Does it make sense to rename this to `browser.web_vital.inp.element`?

Somewhat related, I'm planing on adding a `ui.element.selector` attribute to generically capture selectors but this likely doesn't fit here, given the target could also be a component name. Does that make sense to you?

Lms24: actually, on second thought, "target" is probably more fitting for using it interchangeably on selectors or component names. I'll probably go with `ui.element.target` myself

Lms24 (approved):

model/attributes/browser/browser__web_vital__inp__target.json
logaretm: I did some of research on which name to use, I should've added it to the PR 😓 

So we seem to be following the names that the relevant APIs expose, so by doing the same we end up with `target` because that is what the API exposes, also `targetSelector` in the future.

It looks like this now, across all web vitals:

```
LCP  LargestContentfulPaint.element  ->  browser.web_vital.lcp.element
CLS  LayoutShift.sources             ->  browser.web_vital.cls.source.N
INP  PerformanceEventTiming.target   ->  browser.web_vital.inp.target
```

For `ui.element.*` namespace, it mostly belongs to the element timing API which follows the same pattern we outlined above. I don't want to muddy that IMO.

